PR #14, opened 2022-03-30
Add passes into Current Plan list

pawelacio:
- active passes feature was added (now passes will be visible on current plan page)
- few small style alignments

image


src/components/Offer/OfferStyled.js
```diff
@@ -29,7 +29,7 @@ export const StyledOfferBody = styled.div.attrs(() => ({
   `}
 `;
 
-export const SubscriptionCardWrapperStyled = styled.section.attrs(() => ({
+export const OfferCardWrapperStyled = styled.section.attrs(() => ({
   className: 'msd__subscription-card'
```
iwonakulacz, line 33: we could change it to `msd__offer-card`

src/components/OfferCard/OfferCard.js
```diff
@@ -32,8 +32,7 @@ const SubscriptionCard = ({
   paymentMethod,
   t
 }) => {
-  const isSubscription =
-    getData('CLEENG_OFFER_TYPE') === 'S' || isSubscriptionOffer;
+  const isSubscription = offerType === 'S' || isSubscriptionOffer;
```
iwonakulacz, line 35: do we need `isSubscriptionOffer`? Maybe we could base only on offerType?

pawelacio: `isSubscriptionOffer` was removed from the component

src/components/Offer/OfferStyled.js
pawelacio: Changed

mirydz (approved):

PR #23, opened 2022-05-10
Purchase component refactor

iwonakulacz:


src/containers/OfferContainer/OfferWrapper.js
```diff
@@ -0,0 +1,246 @@
...
+
+  const [errorMsg, setErrorMsg] = useState();
+  const [isLoading, setIsLoading] = useState(true);
+
+  const [couponDetails, setCouponDetails] = useState(null);
+
+  const createOrderHandler = () => {
+    createOrder(offerId).then(orderDetailsResponse => {
+      const { errors } = orderDetailsResponse;
+      if (orderDetailsResponse.errors.length) {
```
pawelacio, line 56: Maybe we can use that errors from the previous line `if (errors.length) {` ?

src/containers/OfferContainer/OfferWrapper.js
```diff
@@ -0,0 +1,246 @@
...
+          couponLoading: false,
+          showMessage: true,
+          message: 'Your coupon has been applied!',
+          messageType: MESSAGE_TYPE_SUCCESS
+        });
+      }
+    });
+  };
+
+  useState(() => {
```
pawelacio, line 137: Shouldn't that be `useEffect`?

mirydz (approved):

src/containers/OfferContainer/OfferContainer.js
```diff
@@ -11,239 +10,235 @@ import {
...
   onSuccess: () => {},
   urlProps: {},
   t: k => k
 };
 
-export { OfferContainer as PureOfferContainer };
-
 // export default withTranslation()(labeling()(OfferContainer));
-export default OfferContainer;
+export default OfferWrapper;
```
pawelacio, line 244: If thats OfferWrapper then we also can change filename

iwonakulacz: I prefer to keep the old name 'OfferContainer', so I rename the OfferWrapper to OfferContainer

pawelacio (approved):

PR #30, opened 2022-05-20
Payment methods prop

pawelacio:
- add **availablePaymentMethods** props in Checkout component and Purchase
- update documentation


src/components/Payment/Payment.js
```diff
@@ -35,34 +35,49 @@ class Payment extends Component {
...
+        paymentMethods: availablePaymentMethods
+      });
+      const defaultMethod = availablePaymentMethods.find(
+        method => method.default
+      );
+      if (defaultMethod) {
+        this.setState({ isPaymentFormDisplayed: true });
+        this.choosePaymentMethod(defaultMethod.id, defaultMethod.methodName);
+      }
+      setData('CLEENG_PAYMENT_METHOD_ID', availablePaymentMethods[0].id);
```
iwonakulacz, line 50: if there is a default payment method we should save this one in the local storage, not the first one

src/components/Payment/Payment.js
```diff
@@ -35,34 +35,49 @@ class Payment extends Component {
...
-      if (paymentMethods) {
-        if (!paymentMethods.length) {
-          this.setState({
-            generalError: t('Payment methods are not defined')
-          });
-        } else {
+    const { t, availablePaymentMethods } = this.props;
+    if (availablePaymentMethods && availablePaymentMethods.length) {
+      this.setState({
+        paymentMethods: availablePaymentMethods
```
iwonakulacz, line 41: we should validate the availablePaymentMethods props -> if it has all required data and if the passes paymentMethods are supported. If validation fails, we can throw an error to the console and fetch the payment details from the API. We should still make sure to show only Adyen and PayPal

src/components/Payment/Payment.js
```diff
@@ -188,8 +202,7 @@ class Payment extends Component {
             <MethodsWrapperStyled>
               {paymentMethods.map(
                 method =>
-                  method.methodName !== 'manual' &&
-                  availablePaymentMethods.includes(method.paymentGateway) && (
```
iwonakulacz, line 192: we could remove `method.methodName !== 'manual'` and just keep the validation if padded payment methods are supported

iwonakulacz (approved):

PR #64, opened 2022-08-02
Skip payment method selection if there is only one method available

pawelacio:


src/components/Payment/Payment.js
```diff
@@ -53,6 +53,11 @@ class Payment extends Component {
         this.setState({ isPaymentFormDisplayed: true });
         this.choosePaymentMethod(defaultMethod.id, defaultMethod.methodName);
       }
+      if (validPaymentMethods.length === 1) {
+        const peymentMethod = validPaymentMethods[0];
```
iwonakulacz, line 57: typo: `paymentMethod`

iwonakulacz (approved):

PR #98, opened 2022-10-12
Add skipAvailableDowngradesStep prop

pawelacio:


README.md
```diff
@@ -516,7 +519,10 @@ const customCancellationReasons = [
 ];
 
 <Provider store={store}>
-  <PlanDetails customCancellationReasons={customCancellationReasons} />
```
iwonakulacz, line 519: what about Subscriptions component? Does this component include the Unsubscribe flow?

src/components/UpdateSubscription/Unsubscribe.js
```diff
@@ -83,6 +84,9 @@ const Unsubscribe = ({
         data: { offerData }
       }
     } = store.getState();
+    if (skipAvailableDowngradesStep) {
```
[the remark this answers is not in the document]
pawelacio: I've started with that but then i saw that other conditions in that function was inside { ... } and to be consistent why decided to do it also here.

README.md
pawelacio: I will check that

iwonakulacz (approved):

PR #100, opened 2022-10-19
Refactor Checkbox component

pawelacio:


src/components/Checkbox/Checkbox.test.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import '@testing-library/jest-dom';
+
+import Checkbox from 'components/Checkbox';
+
+describe('Checkbox component', () => {
+  test('should render correctly without props', async () => {
```
iwonakulacz, line 9: does such tests needs to be 'async'?

pawelacio: Good point. In this file only the last one needs to be `async`
Changed in the newest commit

iwonakulacz (approved):

PR #106, opened 2022-10-25
Refactor Input component

pawelacio:
- Refactor Input component
- Add tests

### DIFF
--- a/src/components/Input/Input.js
+++ b/src/components/Input/Input.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import visibility from './icons/visibleBase64';
 import visibilityOff from './icons/unvisibleBase64';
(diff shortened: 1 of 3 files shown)


src/components/Input/Input.js
```diff
@@ -14,83 +14,81 @@ import {
...
+  floatingLabels,
+  reference
+}) => {
+  return (
+    <InputComponentStyled>
+      <InputElementWrapperStyled
+        error={error}
+        passwordStrength={passwordStrength}
+      >
+        {/* {icon && <InputIconStyled>{icon.render()}</InputIconStyled>} */}
```
iwonakulacz, line 41: can this be deleted?

src/components/Input/Input.js
```diff
@@ -14,83 +14,81 @@ import {
...
+        passwordStrength={passwordStrength}
+        id={`${placeholder}-desc`}
+      >
+        {error}
+      </ErrorWrapper>
+    </InputComponentStyled>
+  );
+};
 
 Input.propTypes = {
```
iwonakulacz, line 93: I would put a time on the props definition here. Be as precise as we can witch prop validation here, eg. type should oneOf with defined available values

src/components/Input/Input.js
```diff
@@ -14,83 +14,81 @@ import {
...
+  showVisibilityIcon,
+  handleClickShowPassword,
+  showPassword,
+  passwordStrength,
+  ariaRequired,
+  ariaInvalid,
+  icon,
+  required,
+  floatingLabels,
+  reference
```
iwonakulacz, line 33: worth checking if we are using all of these props when embedding an Input component. Maybe we could remove unused ones

pawelacio: It looks that all of the props are used but maybe in the process of refactoring components like DateInput, EmailInput some will be deleted. 
image

src/components/Input/Input.js
pawelacio: I was thinking about that line but you're right. I will delete it

iwonakulacz (approved):

PR #109, opened 2022-11-10
move packages to devDependencies

frankenthumbs:
it looks like some dev tools are in dependencies which is then being transferred over to apps that are using this repo. Behavior should remain the same


iwonakulacz:
I'm closing this pull request as this was implemented with https://github.com/Cleeng/mediastore-sdk/commit/e92ab7d94153ac7f51ab6fe982b68cec5eba0314

PR #111, opened 2022-11-30
Apple pay

iwonakulacz:


src/components/Adyen/Adyen.js
```diff
@@ -36,72 +42,94 @@ const Adyen = ({
     });
   };
 
-  const createDropInInstance = async (id, sessionData, paymentMethods) => {
+  const onSelect = component => {
```
jkorczakcleeng, line 45:
```suggestion
  const onSelect = ({ data: { paymentMethod: { type } } }) => {
```

src/util/usePreviousHook.js
```diff
@@ -0,0 +1,10 @@
+import { useEffect, useRef } from 'react';
+
+function usePrevious(value) {
+  const ref = useRef();
```
jkorczakcleeng, line 4:
```suggestion
  const { current } = useRef();
```

src/util/useScriptHook.js
```diff
@@ -0,0 +1,18 @@
+import { useEffect } from 'react';
+
+const useScript = url => {
+  useEffect(() => {
+    const script = document.createElement('script');
```
jkorczakcleeng, line 5:
```suggestion
    const { src, async } = document.createElement('script');
```

iwonakulacz: I need this `script` for
```
    document.body.appendChild(script);

    return () => {
      document.body.removeChild(script);
    };
    ```

src/components/Adyen/Adyen.js
iwonakulacz: fixed

src/util/usePreviousHook.js
iwonakulacz: done

jkorczakcleeng (approved):

PR #120, opened 2022-12-13
Legal notes translations

pawelacio:


jkorczakcleeng (approved):

src/components/Payment/Payment.js
```diff
@@ -240,43 +240,72 @@ class Payment extends Component {
...
-              textDecoration: getData('CLEENG_MY_ACCOUNT_URL')
-                ? 'underline'
-                : 'none'
-            }}
-          >
-            your account
-          </a>{' '}
-          and click &apos;Manage Subscription&apos;.
+          {(() => {
+            if (discountApplied && isInTrial) {
```
iwonakulacz, line 255: as the key here is `legal-notes.trial.period-${period}`, should be checked here if `discountApplied`? 
Maybe we should have here:
1. isInTrial,
2. discountApplied,
3. other cases.
Would be more readable

PR #122, opened 2022-12-15
[MSSDK-1154] - fix dependency cycle problem

jkorczakcleeng:


src/util/appConfigHelper.js
```diff
@@ -1,46 +1,17 @@
...
-    : store.dispatch(setDataInRedux({ name, value }));
+export const setData = (name, value) => localStorage.setItem(name, value);
 
-export const removeData = name =>
-  isLocalStorageAvailable()
-    ? localStorage.removeItem(name)
-    : store.dispatch(removeDataFromRedux({ name }));
+export const removeData = name => localStorage.removeItem(name);
 
 export const sendMessage = msg => {
```
iwonakulacz, line 16: we can remove this - the definition and the usage. It was idea for hosted solution

src/util/appConfigHelper.js
```diff
@@ -1,46 +1,17 @@
 /* istanbul ignore file */
```
iwonakulacz, line 1: this also can be removed :)

iwonakulacz (approved):

iwonakulacz (approved):

PR #133, opened 2023-01-05
bump styled-components version

jkorczakcleeng:


src/components/Input/InputStyled.js
```diff
@@ -46,7 +38,7 @@ export const LabelStyled = styled.label.attrs(() => ({
     width: 100%;
     height: 15px;
 
-    background: ${BackgroundColor};
+    background: ${props => props.theme.BackgroundColor};
```
iwonakulacz, line 41: we could simplify all linke like this to: 
${({theme}) => theme.BackgroundColor};

iwonakulacz (approved):

PR #135, opened 2023-01-09
Payment details refactor

pawelacio:
- changes in some svgs, because viewBox was ignored
- design changes in the paymentCard component
Screenshot 2023-01-09 at 11 30 12
Screenshot 2023-01-09 at 11 31 29
Screenshot 2023-01-09 at 11 32 18
Screenshot 2023-01-09 at 11 33 58


src/components/PaymentCard/PaymentCard.js
```diff
@@ -3,88 +3,126 @@ import PropTypes from 'prop-types';
...
+  ) {
+    LogoComponent = CardTypes[paymentMethodSpecificParams.variant]?.icon;
+    methodTitle = CardTypes[paymentMethodSpecificParams.variant]?.title;
+  } else if (paymentMethod === 'paypal') {
+    LogoComponent = CardTypes[paymentMethod]?.icon;
+    methodTitle = CardTypes[paymentMethod]?.title;
+  } else {
+    LogoComponent = CardTypes[details.paymentMethod]?.icon;
+    methodTitle = CardTypes[details.paymentMethod]?.title;
+  }
```
iwonakulacz, line 68: I would change it to something like
```
const specificParams = paymentMethodSpecificParams.variant || paymentMethod || details.paymentMethod;
const {icon: LogoComponent,  title: methodTitle} = CardTypes[specificParams] || {icon: null/<></> //defaultIcon, title: ''};
```
To test if this works

src/components/PaymentCard/PaymentCard.js
```diff
@@ -3,88 +3,126 @@ import PropTypes from 'prop-types';
...
-                  detail: {
-                    paymentMethod
-                  }
-                })
-              );
-            }}
-          >
-            {t('Edit payment info')}
-          </CardEditStyled>
+          <CardHeaderStyled>Payment method</CardHeaderStyled>
```
iwonakulacz, line 74: I would ask designers once again if we need this duplication of header 'Payment method'

src/components/PaymentCard/PaymentCard.js
```diff
@@ -3,88 +3,126 @@ import PropTypes from 'prop-types';
...
+                )}
+              </CardDetailsStyled>
+            </CardInfoStyled>
+            <CardEditStyled
+              onClick={() => {
+                showInnerPopup({
+                  type: POPUP_TYPES.paymentDetails,
+                  data: details
+                });
+                window.dispatchEvent(
```
iwonakulacz, line 109: you can use eventDispatcher here (see `src/util/eventDispatcher.js`)

src/components/Transactions/Transactions.js
```diff
@@ -79,28 +86,41 @@ const Transactions = ({
...
-                      : subItem.paymentMethod}
-                  </SubTitleStyled>
-                </LeftBoxStyled>
-                <RightBoxStyled>
-                  <IdStyled>{subItem.transactionId}</IdStyled>
-                  <DateStyled>{dateFormat(subItem.transactionDate)}</DateStyled>
-                </RightBoxStyled>
-              </InsideWrapperStyled>
-            ))}
+            {transactions.map(subItem => {
```
jkorczakcleeng, line 89: we can use destructuring on subItem like. map({ paymentMethod, transactionId, ... }) =>

src/components/PaymentCard/Payment.const.js
```diff
@@ -1,21 +1,58 @@
...
+    icon: Paypal,
+    title: 'PayPal'
+  },
+  apple: {
+    icon: Apple,
+    title: 'Apple Pay'
+  },
+  android: {
+    icon: Android,
+    title: 'Google Pay'
```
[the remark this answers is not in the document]
pawelacio: Okey. So for all in-apps edit button should be hidden?

src/components/PaymentCard/PaymentCard.js
pawelacio: That won't work because even if we have `paymentMethod === 'applepay'` there is still `paymentMethodSpecificParams.variant` set with the value `'mc_applepay'`which is not represented in the CardTypes.
I think that we can leave as it is. I know its more lines of code but still readable.
What do you think about that?

src/components/PaymentCard/PaymentCard.js
pawelacio: I've asked about that and also about that holder's name for Paypal

src/components/PaymentCard/Payment.const.js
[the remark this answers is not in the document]
iwonakulacz: For all inapps we should trigger a special inner popup 
https://developers.cleeng.com/docs/payment-information#special-cases

src/components/PaymentCard/PaymentCard.js
iwonakulacz: ok, so maybe: 
```
const getSpecificParam = () => {
  if(paymentMethod === 'card') return paymentMethodSpecificParams.variant;
  if(paymentMethod === 'paypal') return paymentMethod;
  return details.paymentMethod;
}
const {icon: LogoComponent,  title: methodTitle} = CardTypes[getSpecificParam()] || {icon: null/<></> //defaultIcon, title: ''};
```
For me the initial implementation is not so readable, that's why I'm looking for different solution :)

src/components/PaymentCard/PaymentCard.js
```diff
@@ -3,88 +3,124 @@ import PropTypes from 'prop-types';
...
+          </CardExpirationStyled>
+        </CardDetailsStyled>
+      </CardInfoStyled>
+      <SkeletonWrapper height={43} width={170} />
+    </CardInfoWrapStyled>
+  </CardStyled>
+);
+
+const getSpecificPaymentMethod = (
+  paymentMethodSpecificParams,
```
jkorczakcleeng, line 51:
```suggestion
  { variant },
```

pawelacio: That required one additional check if data is loaded but you can see that in the last commit

src/components/PaymentCard/PaymentCard.js
```diff
@@ -3,88 +3,123 @@ import PropTypes from 'prop-types';
...
-    paymentMethodSpecificParams &&
-    CardTypesIcons[paymentMethodSpecificParams.variant]
-      ? CardTypesIcons[paymentMethodSpecificParams.variant]
-      : CardTypesIcons[details.paymentMethod] || null;
+
+  const { icon: LogoComponent, title: methodTitle } = isDataLoaded
+    ? CardTypes[
+        getSpecificPaymentMethod(paymentMethodSpecificParams, paymentMethod)
+      ]
+    : { icon: null, title: '' };
```
iwonakulacz, line 63: if we would put getSpecificPaymentMethod into PaymentCard component, we won't need to pass params (like in line 61)

src/components/PaymentCard/PaymentCardStyled.js
```diff
@@ -98,46 +72,56 @@ export const CardExpirationDateStyled = styled.div.attrs(() => ({
...
+
+export const CardDetailsNameWrapStyled = styled.div`
+  display: flex;
+  margin-bottom: 4px;
+`;
+
+export const CardInfoWrapStyled = styled.div`
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
```
iwonakulacz, line 122: do we need a custom class names for these new elements? I think it may be worth to add

iwonakulacz (approved):

PR #136, opened 2023-01-09
fix cleeng logo

jkorczakcleeng:


iwonakulacz:
Screen Shot 2023-01-09 at 1 07 20 PM
works properly

PR #142, opened 2023-01-13
MSSDK-1089 3DS

iwonakulacz:


src/components/PaymentFinalizationPage/PaymentFinalizationPage.js
```diff
@@ -0,0 +1,67 @@
...
+import FailedPaymentPage from 'components/FailedPaymentPage';
+
+const PaymentFinalizationPage = ({ onSuccess }) => {
+  const adyenRedirectResult = new URLSearchParams(window.location.search).get(
+    'redirectResult'
+  );
+  const orderId = new URLSearchParams(window.location.search).get('orderId');
+  const dispatch = useDispatch();
+
+  const submitRedirectResult = async () => {
```
jkorczakcleeng, line 20: do we need async here?

src/containers/PaymentFinalizationHandler/PaymentFinalizationHandler.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { useSelector, useDispatch } from 'react-redux';
+import PaymentFinalizationPage from 'components/PaymentFinalizationPage';
+import { setShouldShowFinalizePaymentComponent } from 'redux/finalizePaymentSlice';
+
+const PaymentFinalizationHandler = Component => {
```
jkorczakcleeng, line 6: maybe, because it is HOC, we could change the name to sth like `withPaymentFinalizationHandler`?

src/components/PaymentFinalizationPage/PaymentFinalizationPage.js
iwonakulacz: that's true. I will remove it also from other similar places

jkorczakcleeng (approved):

jkorczakcleeng (dismissed):

jkorczakcleeng (approved):

PR #153, opened 2023-02-09
Switches list filtering

pawelacio:


src/components/SubscriptionSwitchesList/SubscriptionSwitchesList.js
```diff
@@ -105,10 +106,15 @@ const SubscriptionSwitchesList = ({
       )
     : [];
 
+  // Filter out the pause subscription
+  const availableFiltered =
+    Array.isArray(availableSorted) &&
```
iwonakulacz, line 111: is there really an option that it won't be an array? Can't we just check if it exists `availableSorted?.filter... `?

src/redux/offersSlice.js
```diff
@@ -0,0 +1,52 @@
+import { createSlice, createAsyncThunk } from '@reduxjs/toolkit';
+import { getOffers } from 'api';
+
+const initialState = {
+  offers: [],
+  pauseOffers: [],
+  pauseOfferIDs: [],
```
iwonakulacz, line 7: as there are multiple offers I would change it to 'pauseOffersIDs'

src/redux/offersSlice.js
```diff
@@ -0,0 +1,52 @@
...
+};
+
+export const fetchOffers = createAsyncThunk(
+  'offers/fetchOffers',
+  async thunkAPI => {
+    try {
+      const result = await getOffers();
+      return result;
+    } catch (err) {
+      return thunkAPI.rejectWithValue(err.message);
```
iwonakulacz, line 19:
```
 async ({rejectWithValue}) => {
    try {
      const result = await getOffers();
      return result;
    } catch (err) {
      return rejectWithValue(err.message);
```

src/redux/offersSlice.js
```diff
@@ -0,0 +1,52 @@
...
+      state.loading = true;
+    },
+    [fetchOffers.fulfilled]: (state, { payload }) => {
+      state.loading = false;
+      state.offers = payload;
+      state.pauseOffers = Array.isArray(payload)
+        ? payload.filter(
+            offer => offer.externalProperties.PAUSE_OFFER === 'true'
+          )
+        : [];
```
iwonakulacz, line 38: is there an option that payload is not an array?

iwonakulacz: payload.filter({externalProperties} => externalProperties.PAUSE_OFFER === 'true')

src/redux/offersSlice.js
```diff
@@ -0,0 +1,52 @@
...
+      state.pauseOfferIDs = Array.isArray(payload)
+        ? payload
+            .filter(offer => offer.externalProperties.PAUSE_OFFER === 'true')
```
iwonakulacz, line 41: .filter({externalProperties} => externalProperties.PAUSE_OFFER === 'true')

src/redux/offersSlice.js
pawelacio: This one is not working but if we want this convention we need to do something like this (this first argument is required)
`async (arg, { rejectWithValue }) => {
    try {
      const result = await getOffers();
      return result;
    } catch (err) {
      return rejectWithValue(err.message);
    }
  }`

src/components/SubscriptionSwitchesList/SubscriptionSwitchesList.js
```diff
@@ -107,9 +107,9 @@ const SubscriptionSwitchesList = ({
     : [];
 
   // Filter out the pause subscription
-  const availableFiltered =
-    Array.isArray(availableSorted) &&
-    availableSorted.filter(offer => !pauseOfferIDs.includes(offer.toOfferId));
+  const availableFiltered = availableSorted?.filter(
+    offer => !pauseOffersIDs.includes(offer.toOfferId)
```
iwonakulacz, line 111: hm we could also do 
```
availableSorted?.filter(
    {toOfferId} => !pauseOffersIDs.includes(toOfferId)
```

iwonakulacz (approved):

src/redux/offersSlice.js
iwonakulacz: true :)

PR #155, opened 2023-02-13
MSSDK-1250 - Move paymentDetails to separate slice

izabelapeza:


src/containers/PaymentInfo/PaymentInfo.component.js
```diff
@@ -137,15 +122,12 @@ const PaymentInfoFn = ({
   };
 
   const updatePaymentDetailsSection = () => {
-    setPaymentDetailsLoading(true);
-    fetchPaymentDetails();
+    dispatch(fetchPaymentDetails());
```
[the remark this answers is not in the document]
iwonakulacz: + maybe we would be able to get rid off 'updatePaymentDetailsSection' function

src/components/UpdatePaymentDetailsPopup/UpdatePaymentDetailsPopup.js
```diff
@@ -77,14 +77,17 @@ const UpdatePaymentDetailsPopup = ({ updatePaymentDetailsSection }) => {
...
-  const { paymentDetails } = useSelector(state => state.paymentInfo);
+  const { paymentDetails } = useSelector(state => state.paymentDetails);
   const { selectedPaymentMethod } = useSelector(state => state.paymentMethods);
   const { loading: isFinalizeAddPaymentDetailsLoading } = useSelector(
     state => state.finalizeAddPaymentDetails
   );
   const [isActionHandlingProcessing, setIsActionHandlingProcessing] = useState(
     false
   );
+  const updatePaymentDetailsSection = () => {
```
iwonakulacz, line 88: I see that this updatePaymentDetailsSection is also passed down to DeletePaymentMethod and FinalizeAddPaymentDetails, could we also get rid of this prop? Simply use dispatch directly in DeletePaymentMethod and FinalizeAddPaymentDetails

src/containers/PaymentInfo/PaymentInfo.component.js
```diff
@@ -6,31 +6,26 @@ import labeling from 'containers/labeling';
...
 import SectionHeader from 'components/SectionHeader';
 import Transactions from 'components/Transactions';
-import { getPaymentDetails, listCustomerTransactions } from 'api';
+import { listCustomerTransactions } from 'api';
 import { PropTypes } from 'prop-types';
 import UpdatePaymentDetailsPopup from 'components/UpdatePaymentDetailsPopup';
 import { useSelector, useDispatch } from 'react-redux';
 import GracePeriodError from 'components/GracePeriodError';
-import { init } from 'redux/publisherConfigSlice';
+import { init, init as initPublisherConfig } from 'redux/publisherConfigSlice';
```
iwonakulacz, line 14: duplicated import, should be: 
`import { init as initPublisherConfig } from 'redux/publisherConfigSlice';`

src/containers/PaymentInfo/PaymentInfo.component.js
```diff
@@ -164,26 +129,19 @@ const PaymentInfoFn = ({
...
       ) : (
         <>
           <SectionHeader>{t('Current payment method')}</SectionHeader>
-          <PaymentMethod
-            paymentDetailsLoading={paymentDetailsLoading}
-            activeOrBoundPaymentDetails={activeOrBoundPaymentDetails}
-            showInnerPopup={showInnerPopup}
-            error={paymentDetailsError}
-          />
+          <PaymentMethod />
```
iwonakulacz, line 144: niiiiceee! :)

src/redux/paymentDetailsSlice.js
```diff
@@ -0,0 +1,45 @@
+import { createSlice, createAsyncThunk } from '@reduxjs/toolkit';
+import { getPaymentDetails } from '../api';
+
+const initialState = {
+  paymentDetails: [],
+  paymentDetailsLoading: true,
+  paymentDetailsError: [],
+  activeOrBoundPaymentDetails: []
```
iwonakulacz, line 8: as this is `paymentDetails` slice I would simplify the state to:
```
{
  paymentDetails: [],
  activeOrBoundPaymentDetails: [],
  loading: true,
  error: [],
}
```
I'm also thinking about 'activeOrBoundPaymentDetails' as these are payment methods that should be presented to customer (`paymentDetails` are all payment details returned from API) . If you have any idea for a better name for it - go ahead :)

src/redux/paymentDetailsSlice.js
```diff
@@ -0,0 +1,45 @@
...
+};
+
+export const fetchPaymentDetails = createAsyncThunk(
+  'paymentDetails/paymentDetails',
```
iwonakulacz, line 12: In this slice we won't do anything more with `paymentDetails` so I would simplify the name to just `paymentDetails`

src/redux/paymentDetailsSlice.js
```diff
@@ -0,0 +1,45 @@
...
+  async (_, { rejectWithValue }) => {
```
iwonakulacz, line 13: I like this solution with `_` for the unused parameter. Could you apply this solution also in `submitPaymentWithoutDetails` and `fetchCustomerOffers`?

src/components/PaymentMethod/PaymentMethod.js
```diff
@@ -31,26 +34,26 @@ const PaymentMethod = ({
...
+  }, []);
+
   return paymentDetailsLoading ? (
     <CardsWrapper numberOfItems={1}>
       <PaymentCard isDataLoaded={false} />
     </CardsWrapper>
   ) : (
     <WrapStyled>
-      {error.length !== 0 ? (
+      {paymentDetailsError.length !== 0 ? (
```
iwonakulacz, line 56: just looking into this return function and I think that we could return what we can, as soon as we can, to avoid nesting. So it would be  (starting from line 49): 
```
if(paymentDetailsLoading)
  return (...)
if(paymentDetailsError.length)
  return (...)
//  the component itself

```

src/components/UpdatePaymentDetailsPopup/UpdatePaymentDetailsPopup.js
iwonakulacz: great that we are not passing is down! is it still needed here then? If not, please delete it :)

src/containers/PaymentInfo/PaymentInfo.js
```diff
@@ -0,0 +1,82 @@
...
+import { PropTypes } from 'prop-types';
+import UpdatePaymentDetailsPopup from 'components/UpdatePaymentDetailsPopup';
+import { useSelector, useDispatch } from 'react-redux';
+import GracePeriodError from 'components/GracePeriodError';
+import { init as initPublisherConfig } from 'redux/publisherConfigSlice';
+import withAddPaymentDetailsFinalizationHandler from 'containers/WithAddPaymentDetailsFinalizationHandler';
+import { updatePaymentDetailsPopup } from 'redux/popupSlice';
+import { WrapStyled } from './PaymentInfoStyled';
+
+const PaymentInfoFn = ({
```
iwonakulacz, line 18: PaymentInfoFn was a temporary name. Lets rename it to PaymentInfo

src/containers/PaymentInfo/PaymentInfo.js
```diff
@@ -0,0 +1,82 @@
...
+  t,
+  displayGracePeriodError
+}) => {
+  const dispatch = useDispatch();
+
+  const { adyenConfiguration: adyenConfigurationStore } = useSelector(
+    state => state.publisherConfig
+  );
+
+  const popupManager = useSelector(state => state.popupManager);
```
iwonakulacz, line 29: const {paymentDetails: paymentDetailsPopup} = useSelector(state => state.popupManager);
and then in l:52
paymentDetailsPopup.isOpen ? ....

src/components/Transactions/Transactions.js
```diff
@@ -56,23 +61,42 @@ const TransactionsSkeleton = () => (
...
+    isTransactionListExpanded
+  } = useSelector(state => state.transactions);
+
+  const transactionsToShow = isTransactionListExpanded
+    ? transactionsList
+    : transactionsList.slice(0, DEFAULT_TRANSACTIONS_NUMBER);
+
+  const dispatch = useDispatch();
+
+  const toggleTransactionsList = () => {
```
iwonakulacz, line 79:
```
  const toggleTransactionsList = () => dispatch(setIsTransactionListExpanded(!isTransactionListExpanded))
```

src/components/Transactions/Transactions.js
```diff
@@ -56,23 +61,42 @@ const TransactionsSkeleton = () => (
...
+      dispatch(setIsTransactionListExpanded(true));
+    }
+  };
+
+  useEffect(() => {
+    if (transactionsList?.length === 0) {
+      dispatch(fetchListCustomerTransactions());
+    }
+  }, []);
+
```
iwonakulacz, line 92: if(isTransactionsSectionLoading)
  return <TransactionsSkeleton />

return (... ) //regular component

With this solution, we will skip the nesting

src/components/Transactions/Transactions.js
```diff
@@ -123,20 +147,19 @@ const Transactions = ({
...
           </TransactionListStyled>
-          {!isShowMoreButtonHidden && (
+          {showToggleButton && (
             <Button
               theme="primary"
               margin="20px 0 0 auto"
               width="unset"
-              label={(isExpanded && t('Show less')) || t('Show more')}
+              label={
+                (isTransactionListExpanded && t('Show less')) || t('Show more')
```
iwonakulacz, line 156: I think it would be better to change it to: `isTransactionListExpanded ? t('Show less')) : t('Show more')`

src/components/Transactions/Transactions.js
```diff
@@ -123,20 +147,19 @@ const Transactions = ({
...
               padding="12px 33px 12px 20px"
             >
-              <ButtonTextStyled isExpanded={isExpanded}>
-                {(isTransactionsItemsLoading && (
-                  <Loader buttonLoader color="#ffffff" />
-                )) ||
-                  (isExpanded && t('Show less')) ||
+              <ButtonTextStyled isExpanded={isTransactionListExpanded}>
+                {(isTransactionListExpanded && t('Show less')) ||
                   t('Show more')}
```
iwonakulacz, line 163: same here: ` isTransactionListExpanded ? t('Show less')) : t('Show more')`

src/redux/transactionsSlice.js
```diff
@@ -0,0 +1,52 @@
...
+
+export const DEFAULT_TRANSACTIONS_NUMBER = 3;
+
+const initialState = {
+  transactionsList: [],
+  transactionsError: [],
+  showToggleButton: false,
+  isTransactionsSectionLoading: false,
+  isTransactionListExpanded: false
+};
```
iwonakulacz, line 12: we could simplify it a bit
```
const initialState = {
  transactions: [],
  error: [],
  loading: false, // as there is no more 'button loading",
  showToggleButton: false,
  isListExpanded: false
};
```

src/redux/transactionsSlice.js
```diff
@@ -0,0 +1,52 @@
...
+
+export const fetchListCustomerTransactions = createAsyncThunk(
+  'transactions/listCustomerTransactions',
+  async (_, { rejectWithValue }) => {
+    try {
+      const responseData = await listCustomerTransactions();
+      return responseData;
```
iwonakulacz, line 19: as we always will use 'items', I would destructure it here:
```
const {items} = await listCustomerTransactions();
return items;
```

src/redux/transactionsSlice.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+);
+
+export const transactionsSlice = createSlice({
+  name: 'transactions',
+  initialState,
+  reducers: {
+    setIsTransactionListExpanded(state, { payload }) {
+      state.isTransactionListExpanded = payload;
+    }
```
iwonakulacz, line 32: I see that we are using this function only to toggle the list, so we could also define it here, not in Transactions component. So more less it would be like:
```
toggleTransactionList(state, {payload}) {
 state. isTransactionListExpanded = !state.isTransactionListExpanded
} 
```
and then in Transactions component `toggleTransactionsList` could be replaced with this one

src/components/Transactions/Transactions.js
```diff
@@ -145,26 +168,14 @@ const Transactions = ({
...
-  error: [],
-  isTransactionsItemsLoading: false,
-  isExpanded: false,
-  t: k => k,
-  isShowMoreButtonHidden: false,
-  isTransactionsSectionLoading: false
+  t: k => k
 };
 
 export { Transactions as PureTransactions };
```
iwonakulacz, line 181: `Transactions` component is also used as a standalone component. Please see the `package.js` file and see how Transactions are imported there (unfortunately its a duplication). Maybe we could get rid of this TransactionList.container and simply use this refactored component?

src/components/Transactions/Transactions.js
izabelapeza: I created separate task for refactor TransactionList, but I can delete it and get rid of TransactionList.container as part of this refactor. It seems to me that the only difference between the TransactionList and Transactions components is the header in the TransactionList, but I can create prop for that.

src/containers/PaymentInfo/PaymentInfo.test.js
```diff
@@ -0,0 +1,86 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import { Provider } from 'react-redux';
+import configureStore from 'redux-mock-store';
+import thunk from 'redux-thunk';
+import PaymentInfo from 'containers/PaymentInfo';
+import 'i18NextInit';
+// eslint-disable-next-line no-unused-vars, import/no-extraneous-dependencies
+import fetchMock from 'jest-fetch-mock';
```
iwonakulacz, line 10: fetchMock is not used anywhere 
do we need this eslint-diable rules?

src/components/Transactions/Transactions.js
```diff
@@ -56,115 +61,120 @@ const TransactionsSkeleton = () => (
...
-  ) : (
-    <WrapStyled>
-      {error.length !== 0 ? (
+const Transactions = ({ t }) => {
+  const {
+    transactions: transactionsList,
+    showToggleButton,
+    error: transactionsError,
+    loading: isTransactionsSectionLoading,
+    isListExpanded: isTransactionListExpanded
```
iwonakulacz, line 70: there is no other list than 'transactionList' so I don't see a point to rename it. Same for other state keys

src/components/Transactions/Transactions.js
```diff
@@ -56,115 +61,120 @@ const TransactionsSkeleton = () => (
...
+  } = useSelector(state => state.transactions);
+
+  const transactionsToShow = isTransactionListExpanded
+    ? transactionsList
+    : transactionsList.slice(0, DEFAULT_TRANSACTIONS_NUMBER);
+
+  const dispatch = useDispatch();
+
+  const toggleTransactionsList = () => dispatch(toggleTransactionList());
```
iwonakulacz, line 79: As this function is used only in one place we can remove it and simply use dispatch(toggleTransactionList()) in Transactions.js l:159

src/components/PaymentMethod/PaymentMethod.js
```diff
@@ -1,25 +1,28 @@
...
-}) => {
+const PaymentMethod = () => {
   const { t } = useTranslation();
   const dispatch = useDispatch();
+
+  const {
+    paymentDetails,
+    error: paymentDetailsError,
+    loading: paymentDetailsLoading,
+    activeOrBoundPaymentDetails
```
iwonakulacz, line 23: I don't see a point to rename 'error' and 'loading' here

src/redux/paymentInfo.js
```diff
@@ -1,9 +1,6 @@
 /* eslint-disable no-param-reassign */
 import { createAction, createReducer } from '@reduxjs/toolkit';
 
-export const SET_PAYMENT_DETAILS = 'SET_PAYMENT_DETAILS';
-export const setPaymentDetails = createAction(SET_PAYMENT_DETAILS);
-
```
iwonakulacz, line 6: the whole file can be deleted

iwonakulacz (approved):

PR #157, opened 2023-02-16
Add subscription pause cancellation flow

pawelacio:


src/components/CancelPausePopup/CancelPausePopup.js
```diff
@@ -0,0 +1,164 @@
...
+  updateList,
+  setSwitchDetails,
+  t
+}) => {
+  const [isLoading, setIsLoading] = useState(false);
+  const [isError, setIsError] = useState(false);
+  const [step, setStep] = useState(1);
+
+  const planDetailsState = useSelector(state => state.planDetails);
+  const switchDetails = planDetailsState.switchDetails[pendingSwitchId];
```
iwonakulacz, line 31: what if any of these (planDetailsState  or planDetailsState.switchDetails) won't exist?

src/components/CancelPausePopup/CancelPausePopup.js
```diff
@@ -0,0 +1,164 @@
...
+  const eventsPayload = {
+    pendingSwitchId,
+    fromOfferId: switchDetails && switchDetails.fromOfferId,
+    toOfferId: switchDetails && switchDetails.toOfferId
```
iwonakulacz, line 35: Could be simplified to:
```
fromOfferId:  switchDetails?.fromOfferId,
toOfferId: switchDetails?.toOfferId
```

src/components/CancelPausePopup/CancelPausePopup.js
```diff
@@ -0,0 +1,164 @@
...
+  };
+
+  const cancelPause = async () => {
+    window.dispatchEvent(
+      new CustomEvent('MSSDK:cancel-switch-action-triggered', {
+        detail: eventsPayload
+      })
+    );
```
iwonakulacz, line 43: eventDispatcher ;)

src/components/CancelPausePopup/CancelPausePopup.js
```diff
@@ -0,0 +1,164 @@
...
+    setIsLoading(true);
+    try {
+      const resp = await updateSwitch(pendingSwitchId);
```
iwonakulacz, line 46: const {errors} = await updateSwitch(pendingSwitchId);

src/components/CancelPausePopup/CancelPausePopup.js
```diff
@@ -0,0 +1,164 @@
...
+    <InnerPopupWrapper
+      steps={2}
+      popupTitle={t('Cancel pause')}
+      currentStep={step}
+      isError={isError}
+    >
+      {step === 1 && (
+        <>
+          <ContentStyled>
+            <TitleStyled>{t('Cancel pause')}</TitleStyled>
```
iwonakulacz, line 86: let's add a special keys for all texts

pawelacio: Ok. I've added those but we have to remember that adding these keys will force people that are using our package to add translations file because there is no fallback string when t function is used with key
image

src/components/CancelPausePopup/CancelPausePopup.js
pawelacio: I dont think that there is any possibility when those values are not available, because this popup is opened from infobox about pending switch so all information should be in store already

pawelacio: Okey. There is also a way to add default value when t() function is used with key:
https://www.i18next.com/translation-function/essentials#passing-a-default-value

src/components/CancelPausePopup/CancelPausePopup.js
pawelacio: Okey. There is also a way to add default value when t() function is used with key:
https://www.i18next.com/translation-function/essentials#passing-a-default-value

iwonakulacz (approved):

PR #161, opened 2023-02-22
Add resume plan option

pawelacio:
image

### DIFF
--- a/src/components/ResumeSubscriptionPopup/ResumeSubscriptionPopup.js
+++ b/src/components/ResumeSubscriptionPopup/ResumeSubscriptionPopup.js
@@ -1,0 +1,226 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { Trans, useTranslation } from 'react-i18next';
+import formatNumber from 'util/formatNumber';
+
+import { subscriptionSwitch } from 'api';
+import Button from 'components/Button';
+import InnerPopupWrapper from 'components/InnerPopupWrapper';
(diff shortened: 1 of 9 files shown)


src/components/ResumeSubscriptionPopup/ResumeSubscriptionPopup.js
```diff
@@ -0,0 +1,249 @@
...
+        currentStep={STEPS_NUMBERS[step]}
+      >
+        <>
+          <ContentStyled>
+            <ImageWrapper>
+              <Close />
+            </ImageWrapper>
+            <TitleStyled step={step}>
+              {t(
+                'resume-subscription-popup.error-header',
```
iwonakulacz, line 120: as this popup is only for pause I would change the translations key to a unique one for pause. Otherwise, publishers accidentally can break resume subscription popup

src/components/SubscriptionManagement/SubscriptionManagement.js
```diff
@@ -201,7 +202,14 @@ const SubscriptionManagement = ({
               theme="primary"
               onClickFn={event => {
                 event.stopPropagation();
-                console.log('Resume subscription');
+                showInnerPopup({
+                  type: POPUP_TYPES.resumeSubscription,
+                  data: {
+                    offerData: {
+                      ...switchSettings[subscription?.offerId].unavailable[0]
```
iwonakulacz, line 209: should be `available`

iwonakulacz (approved):

PR #165, opened 2023-02-24
Mssdk 1066 consents refactor

izabelapeza:


src/components/Consents/Consents.test.js
```diff
@@ -0,0 +1,89 @@
...
+    loading,
+    error
+  },
+  publisherConfig: { publisherId: '' }
+});
+
+const middleware = [thunk];
+const mockStore = configureStore(middleware);
+
+describe('PaymentInfo component', () => {
```
iwonakulacz, line 30: Consents component ;)

src/components/Consents/Consents.test.js
```diff
@@ -0,0 +1,89 @@
...
+  test('should render GeneralErrorStyled (if loading = false)', async () => {
```
iwonakulacz, line 31: I I would specify here that 'should render GeneralErrorStyled when publisherId is not given)

src/components/Consents/Consents.test.js
```diff
@@ -0,0 +1,89 @@
...
+    const { getByText } = render(
+      <Provider store={mockStore(store(false, 'noPublisherId'))}>
+        <Consent />
+      </Provider>
+    );
+
+    expect(
+      getByText(
+        /Unable to fetch terms & conditions. Publisher is not recognized/i
+      )
```
iwonakulacz, line 41: I'm not sure if checking the text is ok here. If we will change the copy in the future it will break the test. We could also check if GeneralErrorStyled is rendered

src/components/Consents/Consents.test.js
```diff
@@ -0,0 +1,89 @@
...
+    const { getByTestId } = render(
+      <Provider store={mockStore(store(true))}>
+        <Consent />
+      </Provider>
+    );
+
+    expect(getByTestId('consents__loader')).toBeInTheDocument();
+  });
+  test('should render consents', async () => {
+    const definitions = [
```
iwonakulacz, line 67: I'm not sure if we should stay with 'definitions' naming. This is a resource from our API that describes publisher consents, so I think naming it as 'publisherConsents' would be better (also with file names, slice,s etc)

src/components/RegisterPage/RegisterForm.js
```diff
@@ -188,9 +189,10 @@ class RegisterForm extends Component {
       generalError,
       showPassword,
       disableActionButton,
-      processing
+      processing,
+      consentError
     } = this.state;
-    const { publisherId, t } = this.props;
```
iwonakulacz, line 193: As we no longer need to access publisherId from props we should also remove passing it to RegisterForm from Register.js:42

src/redux/consentsSlice.js
```diff
@@ -0,0 +1,66 @@
...
+            required: element.required
+          };
+        });
+        const labels = payload.responseData.consents.map(
+          element => element.label
+        );
+        const checked = new Array(definitions.length).fill(false);
+        state.definitions = definitions;
+        state.labels = labels;
+        state.checked = checked;
```
iwonakulacz, line 51: I'm wondering if we need a separate array of labels states. We could simply keep the whole object from API (publisher contents resource as a single source of true) and add a 'checked' state to track the temporary state. So it would be (sample): 
```
const initialState = {
  consents: [
  {
     "broadcasterId": 0,
      "name": "terms",
       "version": "1",
       "value": "https://cleeng.com/cleeng-user-agreement",
       "label": "I accept the <a href=\"https://cleeng.com/cleeng-user-agreement\" target=\"_blank\">Terms and Conditions</a> of Cleeng.",
        "required": true
  }],
  checked: [false],
  loading: false,
  error: ''
};
```
The usage of 'labels' in Consents component could be changed from `labels[index]` to simply `label`

src/redux/rootReducer.js
```diff
@@ -16,6 +16,7 @@ import myaccountReducer from './myaccountSlice';
 import planDetailsReducer from './planDetailsSlice';
 import paymentDetailsSliceReducer from './paymentDetailsSlice';
 import transactionsReducer from './transactionsSlice';
+import consentsReducer from './consentsSlice';
```
iwonakulacz, line 19: as mentioned above I would prefer to name it 'publisherConsents' instead od 'consents'. In the future we will also more customer consents here, so would be good to have a clear names for both resources

src/components/Consents/Consents.test.js
izabelapeza: "When publisherId is not given" was added to test's name, but I left "(if loading = false)" to make a distinction between these and next test.

src/redux/publisherConsentsSlice.js
```diff
@@ -0,0 +1,62 @@
...
+  checked: [],
+  loading: false,
+  error: ''
+};
+
+export const fetchPublisherConsents = createAsyncThunk(
+  'publisherConsents',
+  async (publisherId, { rejectWithValue }) => {
+    try {
+      const responseData = await getConsents(publisherId);
```
iwonakulacz, line 15: getConsents should be adjusted in api folder to properly handle errors

iwonakulacz: const {consents}  = await getConsents(publisherId);

src/redux/publisherConsentsSlice.js
```diff
@@ -0,0 +1,62 @@
...
+    setChecked(state, { payload }) {
+      state.checked[payload] = !state.checked[payload];
+    }
+  },
+  extraReducers: {
+    [fetchPublisherConsents.pending]: state => {
+      state.loading = true;
+    },
+    [fetchPublisherConsents.fulfilled]: (state, { payload }) => {
+      if (payload.responseData && payload.responseData.consents) {
```
iwonakulacz, line 36: this if is not needed

src/redux/publisherConsentsSlice.js
```diff
@@ -0,0 +1,62 @@
...
+        const publisherConsents = payload.responseData.consents.map(element => {
+          return {
+            name: element.name,
+            label: element.label,
+            version: element.version,
+            required: element.required
+          };
+        });
```
iwonakulacz, line 44: the whole 'publisherConsents ' could be simplified to state.publisherConsents = payload

src/components/Consents/Consents.js
```diff
@@ -14,89 +18,37 @@ import {
...
-          generalError: 'noPublisherId'
-        });
-        disabledRegisterButton();
-      }
-    } catch (error) {
-      return error;
-    }
-    return false;
+  const validateConsents = () => {
+    onChangeFn(checked, publisherConsents);
```
iwonakulacz, line 33: to double-check the validation: 
Screen Shot 2023-02-27 at 9 53 26 AM
When user clicks the Register button the required fields should be highlighted. Also, required consents should have '*' at the end
Screen Shot 2023-02-27 at 9 55 39 AM

src/redux/publisherConfigSlice.js
```diff
@@ -18,9 +19,16 @@ export const publisherConfigSlice = createSlice({
     },
     updatePaymentMethods: (state, { payload }) => {
       state.paymentMethods = payload;
+    },
+    setPublisherId: (state, { payload }) => {
+      state.publisherId = payload;
```
iwonakulacz, line 24: please pull main branch. Then you can use init function to set publisher Id

iwonakulacz (approved):

PR #173, opened 2023-03-08
Plan details refactor

pawelacio:


src/components/CancelSwitchPopup/CancelSwitchPopup.js
```diff
@@ -170,8 +175,8 @@ const CancelSwitchPopup = ({
             <Button
               theme="confirm"
               onClickFn={() => {
-                updateList();
-                hideInnerPopup();
+                dispatch(hidePopup());
+                dispatch(updateList());
```
jkorczakcleeng, line 179: could we make one action that do both?

src/components/PauseSubscriptionPopup/PauseSubscriptionPopup.js
```diff
@@ -86,8 +88,8 @@ const PauseSubscriptionPopup = ({
   };
 
   const closePopupAndRefresh = () => {
-    hideInnerPopup();
-    updateList();
+    dispatch(hidePopup());
+    dispatch(updateList());
```
jkorczakcleeng, line 92: same as above - one action? ;)

src/components/SubscriptionSwitchesList/SubscriptionSwitchesList.js
```diff
@@ -12,38 +12,51 @@ import {
...
+    error: isAllSwitchSettingsError
+  } = useSelector(state => state.plan.switchSettings);
+
+  const isOfferSelected = !!offerToSwitch.offerId;
+  const switchSettings = allSwitchSettings[offerToSwitch.offerId];
+  const fromOfferId = offerToSwitch.offerId;
+  const pendingSwtichesToOfferIdsArray = Object.keys(switchDetails).map(
+    item => {
+      return switchDetails[item].toOfferId;
+    }
```
jkorczakcleeng, line 43:
```
item => switchDetails[item].toOfferId;

src/components/CancelSwitchPopup/CancelSwitchPopup.js
pawelacio: hidePopup is action from `popupSlice` and updateList is from the `planDetailsSlice` so dont know if we can/should combine them in one action. What do you think?

src/components/PauseSubscriptionPopup/PauseSubscriptionPopup.js
pawelacio: hidePopup is action from `popupSlice` and updateList is from the `planDetailsSlice` so dont know if we can/should combine them in one action. What do you think?

jkorczakcleeng: make sens :D lets keep as it is ;)

jkorczakcleeng (approved):

PR #177, opened 2023-03-09
MSSDK 1280 Migrate payment's components

izabelapeza:


iwonakulacz (approved):

src/components/Payment/LegalNote/LegalNote.tsx
```diff
@@ -37,10 +36,12 @@ const LegalNote = ({
             return (
               <Trans i18nKey={`legal-notes.trial.period-${period}`}>
                 <strong>
-                  After any free trial and/or promotional period, you will be
-                  charged {{ readablePrice }}
-                  {{ readablePeriod }} or the then-current price plus applicable
-                  taxes on a recurring basis.
+                  <>
```
jkorczakcleeng, line 39: why we need react fragment here?

izabelapeza: Without it I had an error: "This JSX tag's 'children' prop expects a single child of type 'ReactNode', but multiple children were provided." on <strong> tag

PR #178, opened 2023-03-10
MSSDK 1260 Migrate to TypeScript

izabelapeza:


src/api/Customer/getCustomerOffers.ts
```diff
@@ -0,0 +1,49 @@
...
+  offerType: string;
+  paymentGateway: PaymentGateway;
+  paymentMethod: string;
+  pendingSwitchId: unknown;
+  period: string;
+  startedAt: number;
+  status: string;
+  subscriptionId: number;
+  totalPrice: number;
+};
```
iwonakulacz, line 25: getCustomerOffers returns also passes which doesn't have all these fields. Worth to test and add also passes types

src/api/Customer/getPaymentDetails.ts
```diff
@@ -0,0 +1,49 @@
...
+
+type PaymentDetail = {
+  active: boolean;
+  bound: boolean;
+  customerId: number;
+  id: number;
+  paymentGateway: PaymentGateway;
+  paymentMethod: string;
+  paymentMethodId: number;
+  paymentMethodSpecificParams: {
```
iwonakulacz, line 15: not all payment methods will have such paymentMethodSpecificParams object. Eg. paypal will have only 'payerId' and 'holderName'

src/api/Publisher/getPaymentMethods.ts
```diff
@@ -1,7 +1,16 @@
 import fetchWithJWT from 'util/fetchHelper';
 import getApiURL from 'util/environmentHelper';
 
-const getPaymentMethods = () => {
+const getPaymentMethods = (): Promise<{
+  responseData: {
+    paymentMethods: {
+      methodName: string;
+      id: string;
+      [key: string]: unknown;
```
iwonakulacz, line 9: I'm curious, why did you add this line this way? `[key: string]: unknown;` 
In paymentMethods we have also logoUrl and paymentGateway, nothing more

src/components/Offer/Offer.tsx
```diff
@@ -0,0 +1,75 @@
...
+  StyledOfferBody,
+  StyledOfferWrapper,
+  StyledOfferDetailsAndCoupon,
+  StyledOfferCouponWrapper,
+  OfferCardWrapperStyled
+} from './OfferStyled';
+import OfferCheckoutCard from '../OfferCheckoutCard';
+import { OfferProps } from './Offer.types';
+
+const Offer = ({ onSubmit, onPaymentComplete, t }: OfferProps) => {
```
iwonakulacz, line 24: I would rename onSubmit to more like 'onCouponSubmit', so it's only applicable to coupon submittion

src/components/Offer/Offer.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const [coupon, setCoupon] = useState('');
+  const { trialAvailable } = useAppSelector(selectOffer).offer;
```
iwonakulacz, line 26: shouldn't selectOffer return a offer object? They we would have simply `useAppSelector(selectOffer)`

src/components/OfferCheckoutCard/OfferCheckoutCard.tsx
```diff
@@ -19,7 +25,10 @@ import {
 } from './OfferCheckoutCardStyled';
 import Price from '../Price';
 
-const OfferCheckoutCard = ({ isDataLoaded, t }) => {
+const OfferCheckoutCard = ({
+  isDataLoaded = true,
```
iwonakulacz, line 29: just wonder if this is still needed. We could check if offer was fetched (directly from store), and then we could remove this prop

src/redux/orderSlice.ts
```diff
@@ -0,0 +1,252 @@
...
+import { RootState } from './rootReducer';
+
+type Order = {
+  billingAddress: unknown;
+  country: string;
+  couponId: unknown;
+  currency: CurrencyFormat;
+  customer: { locale: string; email: string };
+  customerId: number;
+  discount: { applied: boolean; type: string; periods: number };
```
iwonakulacz, line 14: I think that we could add: type: 'coupon' | 'trial' (not sure if that's all options)

src/redux/popupSlice.ts
```diff
@@ -0,0 +1,68 @@
...
+    ) {
+      state.paymentDetails = { ...state.paymentDetails, ...action.payload };
+    },
+    resetPaymentDetailsPopupState(state) {
+      state.paymentDetails = initialState.paymentDetails;
+    }
+  }
+});
+
+export const selectPaymentDetails = (state: RootState) => state.popupManager.paymentDetails;
```
iwonakulacz, line 61: I would rename to selectPaymentDetails**Popup**

iwonakulacz (dismissed):

iwonakulacz (approved):

PR #189, opened 2023-03-23
MSSDK-1275 fix price rounding and add DKK in planHelper

kkarasek:
## Changes
- set grossPrice as offer's customerPriceIncludingTax and format using formatNumber() 
- changed grossPrice scope to global 
- updated subscription descriptions accordingly
- removed already formatted value in Price component
- fixed total amount formatting in ThankYouPage
- added correct DKK currency display in planHelper

offer_pricebox_rounded

thank_you_page_rounded


jkorczakcleeng (dismissed):

src/components/OfferCheckoutCard/OfferCheckoutCard.js
```diff
@@ -28,34 +28,30 @@ const OfferCheckoutCard = ({ isDataLoaded, t }) => {
     freePeriods,
     freeDays,
     expiresAt,
-    startTime
+    startTime,
+    customerPriceInclTax: grossPrice
```
iwonakulacz, line 32: let's take price from order - it's our source of truth.
So lets keep 
```
const {
    priceBreakdown: { offerPrice },
    taxRate,
    country,
    currency
  } = useSelector(state => state.order.order);
  ```
Tax rate is used when we don't have a gross price (eg. trial), bo let's calculate it based on order - as it was

File: src/components/OfferCheckoutCard/OfferCheckoutCard.js [hunk 2]
```diff
@@ -28,34 +28,30 @@ const OfferCheckoutCard = ({ isDataLoaded, t }) => {
...
-  const {
-    priceBreakdown: { offerPrice },
-    taxRate,
-    country,
-    currency
-  } = useSelector(state => state.order.order);
+  const { country, currency } = useSelector(state => state.order.order);
   const offerType = offerId?.charAt(0);
   const currencySymbol = currencyFormat[currency];
+  const formattedGrossPrice = formatNumber(grossPrice);
```
iwonakulacz, line 37: As we will go back to using an order price value, I would move the old `grossPrice` from l:42 to l:37. Then we will be able to keep the same variable name (`grossPrice`).
Then in the translations, we will have the same variables name, which is crucial for publishers that are already using the translations feature (so in l:41 we want to keep 'grossPrice' as a variable)

src/components/ThankYouPage/ThankYouPage.js
```diff
@@ -31,6 +32,7 @@ const ThankYouPage = ({ onSuccess, t }) => {
   };
   const paymentMethodName = readablePaymentMethod[paymentMethod];
   const currencySymbol = currencyFormat[currency];
+  const formattedTotalAmount = formatNumber(totalAmount);
```
iwonakulacz, line 35: same here - we want to keep the same variable name in translation: totalAmount. You can rename it when getting from store and then here you may have something like: 
```
const totalAmount = formatNumber(totalAmountFromStore);
```

src/components/OfferCheckoutCard/OfferCheckoutCard.js
iwonakulacz: The root cause of the bug (6.99 -> 7 EUR) was caused by `formatNumber(offerPrice + taxRate * offerPrice)`
Screen Shot 2023-03-24 at 8 33 30 AM
This calculation is required to present properly the price for offers with trials (then, totalPrice = 0, and taxValue = 0. We need to present the final price WITH tax).
As a simple solution, I would recommend to: 
Ass util as here https://stackoverflow.com/a/11818658 and use it for `grossPrice`

kkarasek:
### Amendments

- add formatTrialPrice and formatTrialTax functions 
- revert translation changes in CheckoutPriceBox component

### Description
customerPriceInclTax used to check whether grossPrice and taxValue should be rounded or truncated. Should work with currency exchange calculations alike.

The problem has been discussed with backend team, however the calculations provided by core does not work in case currency exchange.

jkorczakcleeng (approved):

src/components/CheckoutPriceBox/CheckoutPriceBox.js
```diff
@@ -66,7 +67,7 @@ const CheckoutPriceBox = ({ t }) => {
               {!taxValue && taxRate && isCouponApplied && (
                 <p style={{ textDecoration: 'line-through' }}>
                   {currencyFormat[currency]}{' '}
-                  {formatNumber(taxRate * offerPrice)}
+                  {formatTrialTax(offerPrice, taxRate, customerPriceInclTax)}
```
iwonakulacz, line 70: rename: `formatTrialTax`, eg. `calculateTaxForFreeOffer`

src/util/formatTrialTax.js
```diff
@@ -0,0 +1,11 @@
+const formatTrialTax = (price, tax, incTax) => {
+  const taxValue = price * tax;
+  const isGrossAndTaxEqual = incTax === Number(price + tax * price).toFixed(2);
```
iwonakulacz, line 3: let's rename isGrossAndTaxEqual to something like `shouldRoundTaxValue`
+ 
`const truncatedTaxValue = (Math.trunc(taxValue * 100) / 100).toFixed(2)`;
return truncatedTaxValue;

src/util/formatTrialPrice.js
```diff
@@ -0,0 +1,11 @@
+const formatTrialPrice = (price, tax, incTax) => {
```
iwonakulacz, line 1: - let's also rename `formatTrialPrice` to something like `calculateGrossPriceForFreeOffer`
- rename params to: netPrice, taxRate, customerPriceInclTax

src/util/formatTrialPrice.js
```diff
@@ -0,0 +1,11 @@
...
+  const grossPrice = price + tax * price;
+  const isGrossAndTaxEqual = incTax === Number(grossPrice).toFixed(2);
```
iwonakulacz, line 3: `shouldRoundGrossPrice`

src/util/formatTrialPrice.js
```diff
@@ -0,0 +1,11 @@
...
+
+  if (isGrossAndTaxEqual) {
+    return grossPrice.toFixed(2);
+  }
+  return (Math.trunc(grossPrice * 100) / 100).toFixed(2);
```
iwonakulacz, line 8: // required to handle price edge cases

iwonakulacz (approved):

iwonakulacz (approved):

PR #202, opened 2023-04-06
Replace expiresAt dates for seasonal offers

pawelacio:


src/components/CancelSwitchPopup/CancelSwitchPopup.js
```diff
@@ -116,8 +117,10 @@ const CancelSwitchPopup = ({
                 {
                   switchDirection,
                   switchOfferTitle,
-                  baseOfferExpirationDate,
-                  baseOfferPrice
+                  baseOfferExpirationDate:
+                    baseOfferExpirationDate === 2145913200
```
jkorczakcleeng, line 121: Could we create const and store it in one place? Sth like:
```
const INFINITE_DATE = 2145913200;

Comment (jkorczakcleeng) [reply src/components/CancelSwitchPopup/CancelSwitchPopup.js:121]: or maybe:
```
const formatExpiresAtText = (date) => {
  const INFINITE_DATE = 2145913200;
  if (date === INFINITE_DATE) {
    return t("the next season start");
  }
  return dateFormat(date);
};

pawelacio: Ive decided to store that in the variable because in one place the text was a little different.

jkorczakcleeng (dismissed):

iwonakulacz (approved):

PR #203, opened 2023-04-06
MSSDK-1332 fix return urls for MyAccount and Checkout

kkarasek:
#### Changes

- make queryParamsString's scope global
- fix url params for adyenConfiguration's myaccountReturnUrl and checkoutReturnUrl

#### Snapshots

Screenshot 2023-04-06 at 14 29 41
Screenshot 2023-04-06 at 14 24 35

[MSSDK-1332]: https://cleeng.atlassian.net/browse/MSSDK-1332?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/util/returnUrlHelper.js
+++ b/src/util/returnUrlHelper.js
@@ -12,13 +12,9 @@ const generateReturnUrl = ({ queryParams, isMyAccount }) => {
     return `${url}${new URL(url).search ? '&' : '?'}${paramsString}`;
   };
 
-  if (isMyAccount && myaccountReturnUrl) {
-    return formatUrl(myaccountReturnUrl);
-  }
+  if (isMyAccount && myaccountReturnUrl) return formatUrl(myaccountReturnUrl);
 
-  if (!isMyAccount && checkoutReturnUrl) {
-    return formatUrl(checkoutReturnUrl);
-  }
+  if (!isMyAccount && checkoutReturnUrl) return formatUrl(checkoutReturnUrl);
 
   if (!queryParams) return window.location.href;
 
--- a/src/util/returnUrlHelper.js
+++ b/src/util/returnUrlHelper.js
@@ -4,18 +4,32 @@ const generateReturnUrl = ({ queryParams, isMyAccount }) => {
   const {
     publisherConfig: { adyenConfiguration }
   } = store.getState();
+  let queryParamsString;
 
-  if (isMyAccount && adyenConfiguration?.myaccountReturnUrl)
-    return adyenConfiguration.myaccountReturnUrl;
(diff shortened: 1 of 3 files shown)


src/util/returnUrlHelper.js
```diff
@@ -4,18 +4,32 @@ const generateReturnUrl = ({ queryParams, isMyAccount }) => {
   const {
     publisherConfig: { adyenConfiguration }
   } = store.getState();
+  let queryParamsString;
 
-  if (isMyAccount && adyenConfiguration?.myaccountReturnUrl)
-    return adyenConfiguration.myaccountReturnUrl;
+  if (isMyAccount && adyenConfiguration?.myaccountReturnUrl) {
+    if (queryParams) {
```
kkarasek, line 10: I was trying to use optional chaining or short circuiting as an alternative here, but unsuccessful - hence if statement. 

If anyone had better idea, that would be appreciated

src/util/returnUrlHelper.js
```diff
@@ -4,18 +4,32 @@ const generateReturnUrl = ({ queryParams, isMyAccount }) => {
...
+      queryParamsString = Object.keys(queryParams).map(
+        key => `${key}=${queryParams[key]}`
+      );
+    }
+    return `${adyenConfiguration.myaccountReturnUrl}?${queryParamsString}`;
```
iwonakulacz, line 15: what in case myaccountReturnUrl (or checkoutReturnUrl) will already have query params? Then we will end up with something like `https://client-page.com/checkout?somecustomparam=123?orderId=12345678`, while we should have  `https://client-page.com/checkout?somecustomparam=123&orderId=12345678`. We should detect there if we should pass '?' or '&'

src/util/returnUrlHelper.js
```diff
@@ -4,18 +4,32 @@ const generateReturnUrl = ({ queryParams, isMyAccount }) => {
...
   } = store.getState();
+  let queryParamsString;
 
-  if (isMyAccount && adyenConfiguration?.myaccountReturnUrl)
-    return adyenConfiguration.myaccountReturnUrl;
+  if (isMyAccount && adyenConfiguration?.myaccountReturnUrl) {
+    if (queryParams) {
+      queryParamsString = Object.keys(queryParams).map(
+        key => `${key}=${queryParams[key]}`
+      );
```
iwonakulacz, line 13: why we skipped the .join here? could we move this logic to a separate function and simply reuse it?

jkorczakcleeng (approved):

iwonakulacz (approved):

PR #218, opened 2023-04-19
MSSDK-1336 Filter payment methods in components

m4tewoosh:
### Description 
https://cleeng.atlassian.net/browse/MSSDK-1336?atlOrigin=eyJpIjoiODNlZWI1M2IxMmI0NDZkMWJjMzFhYjBhOTY1MGExM2EiLCJwIjoiaiJ9

### Updates

added `setAvailablePaymentMethods` to Config (array of methods that we want to display in DropIn)
changed naming (changed also in connector) for sessions:
`bank` -> `zeroPaymentNotSupported`
`standard` -> `zeroPaymentSupported`

added logic for showing PayPal, connector is not accepting array with only ['paypal'] so it's handled on frontend

### DIFF
--- a/src/components/Adyen/Adyen.js
+++ b/src/components/Adyen/Adyen.js
@@ -6,7 +6,11 @@ import labeling from 'containers/labeling';
 import AdyenCheckout from '@adyen/adyen-web';
 import createPaymentSession from 'api/Payment/createPaymentSession';
 import useScript from 'util/useScriptHook';
-import { getAvailablePaymentMethods } from 'util/paymentMethodHelper';
+import {
+  getAvailablePaymentMethods,
+  bankPaymentMethods,
+  bankPaymentMethodsMapper
+} from 'util/paymentMethodHelper';
 import { useSelector } from 'react-redux';
 import Checkbox from 'components/Checkbox';
 import AdyenStyled from './AdyenStyled';
@@ -19,8 +23,6 @@ import {
   getGooglePayEnv
 } from './util/getAdyenConfig';
 import defaultAdyenTranslations from './util/defaultAdyenTranslations';
-
-const bankPaymentMethods = ['ideal', 'sofort', 'directEbanking', 'bcmc_mobile'];
 
 const Adyen = ({
   onSubmit,
@@ -134,15 +136,8 @@ const Adyen = ({
     }
   };
 
-  const onSelect = async ({ type }) => {
-    const typeMapper = {
-      bcmc_mobile: 'bancontact_mobile',
-      directEbanking: 'sofort',
-      bcmc: 'bancontact_card'
-    };
-
-    selectPaymentMethod(typeMapper[type] || type);
-  };
+  const onSelect = async ({ type }) =>
+    selectPaymentMethod(bankPaymentMethodsMapper[type] || type);
 
   const mountStandardDropIn = adyenCheckout => {
     if (standardPaymentMethodsRef?.current) {
@@ -337,6 +332,7 @@ const Adyen = ({
       const shouldCreateStandardPaymentSession = availablePaymentMethods.some(
         ({ methodName }) => !bankPaymentMethods.includes(methodName)
       );
+
       if (shouldCreateStandardPaymentSession) {
         createSession('zeroPaymentSupported');
       }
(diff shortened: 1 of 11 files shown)


src/components/Adyen/Adyen.js
```diff
@@ -240,7 +240,7 @@ const Adyen = ({
         return onSubmit(state, component);
       },
       onActionHandled: () => {
-        if (type === 'bank') {
+        if (type === 'zeroPaymentNotSupported') {
```
iwonakulacz, line 243: nice to have: we could move zeroPaymentNotSupported and zeroPaymentSupported to a cont to avoid typos
const ZEROPAYMENTSUPPORTED = 'zeroPaymentSupported';
const ZEROPAYMENTNOTSUPPORTED = 'zeroPaymentNotSupported';

src/components/Adyen/Adyen.js
```diff
@@ -302,29 +301,39 @@ const Adyen = ({
...
   const generateDropIns = () => {
-    const { paymentMethods } = publisherConfig;
+    const availablePaymentMethods = getAvailablePaymentMethods(
+      publisherPaymentMethods
+    );
+
+    if (!availablePaymentMethods.length) {
+      // use function below to handle case when session returns error -
+      // e.g publisher has only iDeal but you are not using VPN
+      setGeneralError(t('Payment methods not available'));
```
iwonakulacz, line 318: this general error probably won't be needed - to confirm with Ania

src/components/Adyen/Adyen.js
```diff
@@ -302,29 +301,39 @@ const Adyen = ({
...
+      setIsLoading(false);
+      return;
+    }
 
-    if (isMyAccount || totalPrice === 0) {
-      const shouldCreateBankPaymentSession = paymentMethods.some(
+    if (totalPrice === 0) {
```
iwonakulacz, line 323: did you check the myaccount case? there the totalPrice is unavailable but we still wants to present bank payment methods there

src/components/Adyen/Adyen.js
iwonakulacz: or as we are using 'bank' and 'standard' in code we could do: 
const STANDARD_PAYMENT_METHODS = 'zeroPaymentSupported';
const BANK_PAYMENT_METHODS = 'zeroPaymentNotSupported';

src/util/appConfigHelper.js
```diff
@@ -163,5 +173,7 @@ export default {
   setOfferSelectionUrl,
   setJWT,
   setRefreshToken,
-  setTermsUrl
+  setTermsUrl,
+  setHidePayPal,
+  setVisibleAdyenPaymentMethods
```
iwonakulacz, line 178: please update READM.me with these two new Config methods

src/components/Adyen/Adyen.js
iwonakulacz: If this is the case when NO payment methods are available at all - then we probably should keep this error message

src/util/paymentMethodHelper.js
```diff
@@ -72,3 +73,19 @@ export const validatePaymentMethods = (
...
+  );
+
+  const availablePaymentMethods = visibleAdyenPaymentMethods.length
+    ? publisherPaymentMethods.filter(({ methodName }) =>
+        visibleAdyenPaymentMethods.includes(methodName)
+      )
+    : publisherPaymentMethods;
+
+  return availablePaymentMethods;
+};
```
iwonakulacz, line 91: I think we should keep these payment methods (and logic in redux). Currently, we have state.publisherConfig.paymentMethods which represents payment methods set in cleeng admin. I think that when publisher sets visible payment methods we could save them in the state.publisherConfig.visiblePaymentMethods. It's an improvement, not a must-have, but worth to validate :)

src/components/Adyen/Adyen.js
m4tewoosh: It's case when there are no payment methods enabled by publisher and also when there are some methods but user is using (in array) only those not enabled

src/components/Adyen/Adyen.js
m4tewoosh: there should be this first condition with `isMyAccount`, my mistake when merging

PR #219, opened 2023-04-19
Add some alignments for seasonal subscriptions

pawelacio:
### Description 

Some modification in current plan and checkout to handle seasonal subscription type

### Screenshots 
image

image


src/components/Price/Price.js
```diff
@@ -18,20 +18,22 @@ const addSpaceAfterNumber = str => {
...
-        <Trans i18nKey="price-additional-label"> </Trans>
-      </AdditionalLabelStyled>
-    </InnerWrapper>
-    {period && (
-      <PeriodStyled>/&nbsp;{addSpaceAfterNumber(period)}</PeriodStyled>
-    )}
-  </WrapperStyled>
-);
+const Price = ({ currency, price, period }) => {
+  return (
```
kjhank, line 22: Why the explicit return?

kjhank (changes requested):

src/components/Price/Price.js
pawelacio: I forgot abut that. Already changed it.

kjhank (dismissed):

PR #223, opened 2023-05-09
MSSDK-1390 Change the copy for bank payment methods in my account

m4tewoosh:
### Description 

Simply change of bank copy in MyAccount


src/components/Adyen/Adyen.js
```diff
@@ -76,7 +76,7 @@ const Adyen = ({
     const isFree = totalPrice === 0;
     const isSubscription = offerId?.charAt(0) === 'S' && !isMyAccount;
```
iwonakulacz, line 77: I think we can remove `!isMyAccount` from here

PR #225, opened 2023-05-10
Update unsubscribe flow

pawelacio:
### Description 

Updating the unsubscribe flow and split the pause/downgrades screens into separate views

### Updates

Object based solution was replaced with Array based becuase its easear to deal with the order of the elements and adding something dynamically.

### Screenshots 
image

### Testing 
not needed

### Additional Notes
not needed


src/components/InnerPopupWrapper/InnerPopupWrapperStyled.js
iwonakulacz, line 1: Copy in the pause popup should be centered vertically

src/components/UpdateSubscription/Unsubscribe.js
```diff
@@ -96,23 +88,52 @@ const Unsubscribe = ({
...
+      const downgradesFiltered = downgrades.filter(
+        ({ toOfferId }) => !pauseOffersIDs.includes(toOfferId)
+      );
+      return !!downgradesFiltered.length;
+    }
+    return false;
+  };
+
+  const shouldShowPauseScreen = () => {
+    const downgrades = getDowngrades();
```
iwonakulacz, line 101: couldn't we just set this `downgrade` value once, outside of these functions?

src/components/InnerPopupWrapper/InnerPopupWrapperStyled.js
pawelacio: To be honest its vertically centered but on the designs we have big space between buttons and that make it look like its not

iwonakulacz (approved):

PR #237, opened 2023-05-23
[MSSDK-1385] Add right-to-left support in mediastore-sdk

pawelacio:
### Description

Add support for dir="rtl" attribute in mediastore-sdk library

### Updates
- adding new helper `isRTL()` that will return true if dir="rtl" is set in the document
- based on the document.dir value add some style alignment (margins, paddings)

### Screenshots
image
image
image
image

image

### DIFF
--- a/src/components/Checkbox/CheckboxStyled.js
+++ b/src/components/Checkbox/CheckboxStyled.js
@@ -1,5 +1,4 @@
 import styled, { css } from 'styled-components';
-import { isRTL } from 'styles/RTLHelper';
 import {
   FontColor,
   ErrorColor,
@@ -40,7 +39,7 @@ export const ConsentDefinitionStyled = styled.div.attrs(props => ({
   }`
 }))`
   position: relative;
-  padding-left: 10px;
+  padding-inline-start: 10px;
   margin-top: 0;
 
   font-weight: 400;
@@ -58,12 +57,6 @@ export const ConsentDefinitionStyled = styled.div.attrs(props => ({
     props.checked &&
     css`
       opacity: 1;
-    `}
-
-  ${isRTL() &&
-    css`
-      padding-left: 0;
-      padding-right: 10px;
     `}
 `;
 
--- a/src/components/Checkbox/CheckboxStyled.js
+++ b/src/components/Checkbox/CheckboxStyled.js
@@ -1,4 +1,5 @@
 import styled, { css } from 'styled-components';
+import { isRTL } from 'styles/RTLHelper';
 import {
   FontColor,
   ErrorColor,
@@ -57,6 +58,12 @@ export const ConsentDefinitionStyled = styled.div.attrs(props => ({
     props.checked &&
     css`
       opacity: 1;
+    `}
+
+  ${isRTL() &&
+    css`
+      padding-left: 0;
+      padding-right: 10px;
     `}
 `;
 
--- a/src/components/CouponInput/CouponInputStyled.js
+++ b/src/components/CouponInput/CouponInputStyled.js
@@ -1,6 +1,7 @@
 import styled, { css } from 'styled-components';
 import * as Colors from 'styles/variables';
 import { media } from 'styles/BreakPoints';
+import { isRTL } from 'styles/RTLHelper';
 import { MESSAGE_TYPE_SUCCESS } from 'components/Input/InputConstants';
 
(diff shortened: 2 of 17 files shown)


src/components/Checkbox/CheckboxStyled.js
```diff
@@ -58,6 +59,12 @@ export const ConsentDefinitionStyled = styled.div.attrs(props => ({
     css`
       opacity: 1;
     `}
+
+  ${isRTL() &&
+    css`
+      padding-left: 0;
```
kjhank, line 65: If you use logical properties instead of the cartesian top/right/bottom/left directions, then adding the `lang="rtl"` attribute will flip your styles automatically.
So instead of setting sth like

```
p {
  padding-left: 2em;
  padding-right: 5em;
}

[dir="rtl"] p {
  padding-left: 5em;
  padding-right: 2em;
}
```

you can just do
```
p {
  padding-inline-start: 2em;
  padding-inline-end: 5em;
}
```

or just
```
p {
  padding-inline: 2em 5em;
}
```

and setting `dir="rtl"` on the element takes care of the rest: https://codepen.io/kjhank/pen/XWxOXRr

kjhank (changes requested):

src/components/Checkbox/CheckboxStyled.js
pawelacio: Ok. Good to know. I will try that one.

src/components/CouponInput/CouponInputStyled.js
```diff
@@ -84,6 +85,12 @@ export const InputElementStyled = styled.input.attrs(() => ({
         `}
     `}
 
+  ${isRTL() &&
```
kjhank, line 88: Or you could do the same as you did with the paddings & margins and set the `inset-inline-start` property to `37px` and get rid of the `right`s and `left`s in here and in :74

src/components/CouponInput/CouponInputStyled.js
```diff
@@ -110,6 +117,12 @@ export const CloseButtonStyled = styled.button.attrs(() => ({
   align-items: center;
   cursor: pointer;
 
+  ${isRTL() &&
+    css`
+      right: 7px;
```
kjhank, line 122: as above

src/components/MyAccountInput/MyAccountInputStyled.js
```diff
@@ -39,6 +40,11 @@ export const InputElementStyled = styled.input.attrs(() => ({
   font-size: 13px;
   line-height: 13px;
 
+  ${isRTL() &&
+    css`
+      text-align: right;
```
kjhank, line 45: `text-align: start` and you can get rid of this 😄 https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties/Floating_and_positioning

src/components/Transactions/TransactionsStyled.js
```diff
@@ -62,8 +63,13 @@ export const RightBoxStyled = styled.div`
   flex-direction: column;
   justify-content: space-between;
   flex-shrink: 0;
-  margin-left: 20px;
+  margin-inline-start: 20px;
   text-align: right;
+
+  ${isRTL &&
+    css`
+      text-align: left;
```
kjhank, line 71: Now you can get rid of :67 and this and set `text-align: end`.

src/styles/RTLHelper.js
```diff
@@ -0,0 +1,4 @@
+// eslint-disable-next-line import/prefer-default-export
+export const isRTL = () => {
```
kjhank, line 2: `const isRTL = () => document.dir === 'rtl';`

kjhank (changes requested):

kjhank (dismissed):

src/styles/RTLHelper.js
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line import/prefer-default-export
+export const isRTL = () => document.dir === 'rtl';
```
kjhank, line 2: One more thing, actually. Are we sure the `dir` attribute will be set?
There's also the `Intl` API we could use instead:
```
const intl = new Intl.Locale(navigator.language);
const isRtl = intl.textInfo.direction === 'rtl';
```

m4tewoosh (approved):

PR #258, opened 2023-07-03
[MSSDK-1391] Emit event when the user should be logout

m4tewoosh:
https://cleeng.atlassian.net/browse/MSSDK-1391

### DIFF
--- a/src/components/Payment/Payment.tsx
+++ b/src/components/Payment/Payment.tsx
@@ -34,7 +34,7 @@ import {
 import eventDispatcher, {
   MSSDK_PURCHASE_FAILED,
   MSSDK_PURCHASE_SUCCESSFUL,
-  MSSDK_TOKEN_EXPIRED
+  MSSDK_AUTH_FAILED
 } from '../../util/eventDispatcher';
 import LegalNote from './LegalNote/LegalNote';
 import PayPal from './PayPal/PayPal';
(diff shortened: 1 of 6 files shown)


src/components/Payment/Payment.tsx
```diff
@@ -94,6 +94,7 @@ const Payment = ({ onPaymentComplete }: PaymentProps) => {
         .unwrap()
         .catch(errors => {
           if (errors.includes('JWT')) {
+            window.dispatchEvent(new CustomEvent('MSSDK:token-expired'));
```
iwonakulacz, line 97: we have a dedicated function for that, lets use it: src/uril/eventDispatcher

src/util/fetchHelper.js
```diff
@@ -60,6 +60,7 @@ const fetchWithJWT = async (url, options = {}) => {
   const refreshToken = retrieveRefreshToken();
 
   if (isExpired && !refreshToken) {
+    window.dispatchEvent(new CustomEvent('MSSDK:token-expired'));
     Auth.logout();
```
iwonakulacz, line 64: should be also added in isJWTExpired function - when there is no JWT

src/services/auth.js
```diff
@@ -111,6 +111,7 @@ class Auth {
     const isExpired = now > decoded.exp;
 
     if (isExpired && !refreshToken) {
+      window.dispatchEvent(new CustomEvent('MSSDK:token-expired'));
```
iwonakulacz, line 114: this function (isLogged) is executed in a few places in our app but it doesn't have a whole refresh logic and it's never executed with the API requests. I think it's should not be added here

iwonakulacz (dismissed):

iwonakulacz (approved):

iwonakulacz (approved):

PR #259, opened 2023-07-03
fix current plan and subscriptionSwitchesList - error handling

iwonakulacz:
### Description

Fix error handling after the TS merge to main 

### Updates

- errors default value was changes to `null`. These are following up changes


kkarasek (approved):
2x 👍

PR #260, opened 2023-07-05
move types/react-redux to devDep

iwonakulacz:
### Description

Types packages are used only for development so should be placed in devDependecies

### Updates

- move types/react-redux from dependencies to devdependencies


kkarasek (approved):
👍

PR #270, opened 2023-07-19
fix/MSSDK-1468: CouponInput focus styles and sequential navigation

kkarasek:
### Description

The current implementation of `CouponInput `suffers from incorrect sequential focus navigation, leading to issues while browsing the page using the `tab` key. This becomes evident when users attempt to switch focus from the previous element to the `Redeem coupon` button to access the form. However, the first element that unexpectedly captures the focus is `CloseButton`, causing a suboptimal user experience.

### Updates

To resolve the issue, both `CloseButton` and the inner `Input` elements were encapsulated in conditional rendering. By doing so, these invisible components are being prevented from receiving focus in the DOM. This ensures that users can smoothly switch focus to the 'Redeem coupon' button and open the form without any unexpected detours.

### Screenshots

image

### Additional notes

Appropriate and accessible focus styles to be discussed w/ Design Team

### DIFF
--- a/src/components/CouponInput/CouponInput.tsx
+++ b/src/components/CouponInput/CouponInput.tsx
@@ -1,4 +1,4 @@
-import { KeyboardEvent, useEffect, useState } from 'react';
+import { KeyboardEvent, useEffect, useRef, useState } from 'react';
 import { MESSAGE_TYPE_SUCCESS } from 'components/Input/InputConstants';
 import Loader from 'components/Loader';
 import Button from 'components/Button';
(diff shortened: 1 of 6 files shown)


src/components/CouponInput/CouponInput.tsx
```diff
@@ -64,28 +55,15 @@ const CouponInput = ({
...
-
-  const onRedeemClick = async () => {
-    if (!isOpened) {
+  const handleRedeem = async () => {
+    if (!isOpen) {
       window.dispatchEvent(
         new CustomEvent('MSSDK:redeem-coupon-button-clicked', {
           detail: { source }
         })
       );
```
iwonakulacz, line 64: let's use `eventDispatcher` here (src/util/eventDispatcher.js) :)

kkarasek: done ;)

iwonakulacz (approved):

PR #275, opened 2023-08-03
[MSSDK-1471][MSSDK-1472] As a gifter I would like to have possibility to edit delivery details

m4tewoosh:
### Description
https://cleeng.atlassian.net/browse/MSSDK-1471
https://cleeng.atlassian.net/browse/MSSDK-1472

### Screenshots
image
image


src/components/DeliveryDetails/RecipientForm/RecipientForm.tsx
```diff
@@ -18,7 +18,11 @@ import {
   validateRecipientEmail
 } from './validators';
 
-const RecipientForm = () => {
+type RecipientFormProps = {
+  isMyAccount?: true;
```
iwonakulacz, line 22: `isMyAccount?: boolean;`

src/components/DeliveryDetails/RecipientForm/RecipientForm.tsx
```diff
@@ -18,7 +18,11 @@ import {
...
+};
+
+const RecipientForm = ({ isMyAccount }: RecipientFormProps) => {
```
iwonakulacz, line 25: `const RecipientForm = ({ isMyAccount = false }: RecipientFormProps) => {`

src/components/DeliveryDetails/RecipientForm/RecipientForm.tsx
```diff
@@ -112,14 +116,20 @@ const RecipientForm = () => {
...
+            'recipientForm.info-text',
+            'To edit your gift delivery details, access MyAccount and click on the corresponding transaction.'
+          )}
+        </InfoText>
+      )}
     </StyledRecipientForm>
   );
 };
 
+RecipientForm.defaultProps = {
```
iwonakulacz, line 131: should be deleted - defaultProps are part of prop types which we are not using anymore

src/components/EditDeliveryDetailsPopup/EditDeliveryDetailsPopup.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const [currentStep, setCurrentStep] = useState(1);
+
+  useEffect(() => {
+    return () => {
+      dispatch(resetDeliveryDetailsState());
+    };
+  }, []);
+
+  // get offerTitle from store or props
+  const offerTitle = '1 Month Subscription - Premium Plan';
```
iwonakulacz, line 31: indeed, let's take offer title from the store :)

src/components/EditDeliveryDetailsPopup/EditDeliveryDetailsPopup.tsx
```diff
@@ -0,0 +1,94 @@
...
+                  'edit-delivery-details-popup.thank-you-page.header',
+                  'Delivery Details Updated'
+                )}
+              </ThankYouPageHeaderStyled>
+              <ThankYouPageInfoTextStyled>
+                {t(
+                  'edit-delivery-details-popup.thank-you-page.info-text-1',
+                  'Thank you for updating your delivery details for your'
+                )}
+                <p>{offerTitle}</p>
```
iwonakulacz, line 74: also let's add a translation key to offer title here, key: `offer-title-${offerId}`

src/components/Transactions/Transactions.js
```diff
@@ -1,21 +1,28 @@
 /* eslint-disable no-nested-ternary */
 import { useEffect } from 'react';
```
iwonakulacz, line 2: Could we rewrite this to TS?

src/containers/PaymentInfo/PaymentInfo.tsx
```diff
@@ -58,17 +64,19 @@ const PaymentInfo = ({
...
+                <>
+                  {t(
+                    'paymentinfo.current-payment-method',
+                    'Current payment method'
+                  )}
+                </>
+              </SectionHeader>
+              <PaymentMethod />
+            </section>
+          )}
```
iwonakulacz, line 79: How does it look? Could you add screenshots to the pull request description? Shouldn't we hide the entire section (with payment history) when this popup is open?

src/redux/innerPopupReducer.js
```diff
@@ -8,7 +8,8 @@ export const POPUP_TYPES = {
   switchPlan: 'switchPlan',
   cancelSwitch: 'cancelSwitch',
   cancelPause: 'cancelPause',
-  resumeSubscription: 'resumeSubscription'
+  resumeSubscription: 'resumeSubscription',
+  editDeliveryDetails: 'editDeliveryDetails'
```
iwonakulacz, line 12: do we need it here, as this popup is manage by popupSlice?

src/components/DeliveryDetails/RecipientForm/RecipientForm.tsx
m4tewoosh: It produced eslint error, but now I found that moving RecipientFormProps to separate file and then importing it fixed it

src/containers/PaymentInfo/PaymentInfo.tsx
m4tewoosh: How it looks now:
image

m4tewoosh: And also add title and text to popup first step

src/redux/innerPopupReducer.js
m4tewoosh: nope, it's in `popupSlice.ts` so will delete here

src/components/EditDeliveryDetailsPopup/EditDeliveryDetailsPopup.tsx
```diff
@@ -0,0 +1,110 @@
...
+      <ContentStyled>
+        {currentStep === 1 ? (
+          <>
+            <HeaderStyled>
+              {t('edit-delivery-details-popup.header', 'Edit Delivery Details')}
+            </HeaderStyled>
+            <InfoTextStyled>
+              {t(
+                'edit-delivery-details-popup.info-text-1',
+                'You have selected the issue regarding your'
```
iwonakulacz, line 54: are these copies form the design or Anna? If from design, do we have a confirmation for Anna for all copies?

src/components/Transactions/Transactions.tsx
```diff
@@ -67,12 +76,14 @@ const Transactions = () => {
     error,
     loading,
     isListExpanded
-  } = useSelector(state => state.transactions);
-  const { pauseOffersIDs } = useSelector(state => state.offers);
+  } = useAppSelector(state => state.transactions);
```
iwonakulacz, line 79: for appSelector we were adding a dedicated function to achieve something like: 
`useAppSelector(selectTransactions);` 
See selectFinalizePayment, selectOrder, etc.

src/containers/PaymentInfo/PaymentInfo.tsx
iwonakulacz: I'm still not sure why we need to embed Transactions when `isEditDeliveryDetailsPopupOpened`. Any particular reason?

src/components/Transactions/Transactions.tsx
m4tewoosh: To do it I need to rewrite `offersSlice.js` to TypeScript, I will do it

src/containers/PaymentInfo/PaymentInfo.tsx
m4tewoosh: This Popup is displayed inside `Transactions` component, as I understand we have `TransactionList` component that can be used alone without `PaymentInfo` so this Popup needs to be deeper. Instead of returning solo `Transactions` here in `PaymentInfo` I can return only `EditDeliveryDetailsPopup` but it will do same thing.

src/components/EditDeliveryDetailsPopup/EditDeliveryDetailsPopup.tsx
m4tewoosh: Nope, will confirm then. But in future it would be nice if finished Figma design wouldn't need copy confirmation.

src/components/Transactions/Transactions.tsx
m4tewoosh: added it for offers, using it with transactions produces ts errors because whole transactions types needs to be added and slice rewritten, better do it in other task

src/containers/PaymentInfo/PaymentInfo.tsx
m4tewoosh: Now return only `EditDeliveryDetailsPopup` and made it a bit simplified

src/components/EditDeliveryDetailsPopup/EditDeliveryDetailsPopup.tsx
m4tewoosh: copies confirmed and updated, also will be different when gift is already sent

iwonakulacz (approved):

PR #280, opened 2023-08-11
fix/MSSDK-1511: calculate gross price only when tax rate is not 0

iwonakulacz:
### Description

For some specific cases the wrong price rounding is applied in the checkout. In this fix, for cases where tax rate is 0 I return the net price as gross price.

### Updates

- return net price when tax rate is 0

### Screenshots
Added in the ticket.

### DIFF
--- a/src/util/calculateGrossPriceForFreeOffer.js
+++ b/src/util/calculateGrossPriceForFreeOffer.js
@@ -3,6 +3,7 @@ const calculateGrossPriceForFreeOffer = (
   taxRate,
   customerPriceInclTax
 ) => {
+  if (taxRate === 0) return netPrice;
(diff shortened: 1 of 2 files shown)


src/util/calculateGrossPriceForFreeOffer.js
```diff
@@ -3,6 +3,7 @@ const calculateGrossPriceForFreeOffer = (
   taxRate,
   customerPriceInclTax
 ) => {
+  if (taxRate === 0) return netPrice;
```
kkarasek, line 6: let's maybe add .toFixed(2) here as well

m4tewoosh (approved):

PR #287, opened 2023-08-25
feat/MSSDK-1521-handle-error-when-updating-delivery-details

kkarasek:
### Description

Previously, regardless of whether the delivery details editing process was successful or not, the ThankYouPage always displayed a positive outcome message. 
We needed to handle the error and render relevant information about a failure to address this issue. 

### Updates

- added conditional rendering block to display relevant ThankYouPage content
- updated translations.json file
- changed the 'Back to settings' button's margin to 80px

### Screenshots

Screenshot 2023-08-25 at 18 27 12

Screenshot 2023-08-25 at 18 27 25

Screenshot 2023-08-25 at 18 28 28


src/components/EditDeliveryDetailsPopup/EditDeliveryDetailsPopup.tsx
```diff
@@ -186,29 +190,58 @@ const EditDeliveryDetailsPopup = () => {
...
-              'edit-delivery-details-popup.thank-you-page.info-text-2',
-              'Your changes have been saved and will be reflected in your next delivery.'
-            )}
-          </InfoTextStyled>
-          <Button theme="confirm" onClickFn={() => dispatch(hidePopup())}>
-            {t('edit-delivery-details-popup.button.back', 'Back to settings')}
-          </Button>
-        </ThankYouPageStyled>
+        {error ? (
+          <ThankYouPageStyled>
```
kkarasek, line 194: I had a bit of a naming dilemma here. Not entirely sure if it's best to keep the ThankYouPageStyled as it is, and change only the content based on the error value, or if we should create a new identical div named ErrorPageStyled to avoid any confusion. 

Let me know your thoughts🤔

m4tewoosh: You can try to do something like that in `EditDeliveryPopupStyled.ts`: 
`export const ErrorPageStyled = styled(ThankYouPageStyled);`
I'm not sure if this syntax is correct but you know what I mean.

But this component is a bit complex now so you can also extract `ErrorPageStyled` to solo component in `EditDeliveryDetailsPopup` folder and just import it here, as I see it doesn't use any variables besides `hidePopup` func and translations so it should be fairly easy.

kkarasek: @m4tewoosh changed applied. I've just added ErrorPageStyled that extends the ThankYouPageStyled. Should be ok, I think.

m4tewoosh (approved):

PR #297, opened 2023-09-20
feat/MSSDK-1536: export RedeemGift and provide authentication handler

kkarasek:
### Description

The purpose of this change is to provide publishers with the option to use the 'RedeemGift' component as a separate chunk in their apps. The component should be exportable and must retain all of its original features. Moreover, an additional authentication handler or condition check should be implemented to ensure the safety of gift redemption. This will prevent users who do not have an account from redeeming gifts.

### Updates

- made the RedeemGift exportable
- implemented conditional rendering to check if users are logged in
- added `isNotAuth` error type and `isRedeemGift` prop to the ErrorPage component to conditionally render the header
- added `MSSDK_AUTH_FAILED` event dispatch in case the user is not logged in
- updated the `translations.json` file with the new key

### Screenshots

Screenshot 2023-10-05 at 16 50 42

Screenshot 2023-10-05 at 16 50 08

Screenshot 2023-10-05 at 16 51 12

### Additional Notes
There will be a need to update Cleeng's documentation. Here's the link to the draft:
https://docs.google.com/document/d/1aTvDuYG-E87d4wV1PfIW_dv34bLsNBp8RVqBKJIeJns/edit?usp=sharing

It may be worth considering adding some sort of an error page for all of the Checkout's subcomponents. We don't have any explicit handlers for that currently ✅


src/components/RedeemGift/RedeemGift.tsx
```diff
@@ -94,53 +95,57 @@ const RedeemGift = ({ onBackClick, onSuccess }: RedeemGiftProps) => {
...
-            <Button
-              disabled={!giftCode}
-              theme="confirm"
-              onClickFn={handleVerifyGift}
-            >
-              {isVerifyLoading ? (
-                <Loader buttonLoader color="#ffffff" />
-              ) : (
-                t('redeem-gift.button.verify', 'Verify')
+      {Auth.isLogged() && (
```
jkorczakcleeng, line 98: please avoid nested ternary operator and &&. Use If return instead.

kkarasek: refactored, thx for the tip!

jkorczakcleeng (approved):

PR #299, opened 2023-09-21
feat/MSSDK-1550: Fix session endpoint call

pawelacio:
### Description

During the process of checkout/purchase we are storing the orderId in the local storage. If the user after some time enter again the checkout/purchase we are showing that old order. There is a possibility that the order is invalid or expired and that is causing the 422 /sessions problem.
To remove that problem we made some changes to wait with the /sessions call as long as getOrder/createOrder is taking place. After succesful getting order or order creation the sessions request will be called. 

### Updates

- add async/await to make sure that we will be wainting for the end of the getOrder/createOrder
- because the unsuccesful getOrder is always followed by the createOrder we wont change the loading to false (just to make sure that, in that moment, the Offer component with the session call wont be rendered)
- clear order with the initialState in the orderSlice to make sure that we wont render the Offer component at the next Purchase/Checkout component render

### DIFF
--- a/src/containers/OfferContainer/OfferContainer.tsx
+++ b/src/containers/OfferContainer/OfferContainer.tsx
@@ -18,6 +18,7 @@ import {
   fetchCreateOrder,
   fetchGetOrder,
   fetchUpdateCoupon,
+  clearOrder,
   selectOrder
 } from 'redux/orderSlice';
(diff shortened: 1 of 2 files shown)


kkarasek (approved):

kkarasek:
thumbs up to the description!

PR #304, opened 2023-10-05
fix/MSSDK-1556: change coupon discount copy in offer in trial

jkorczakcleeng:
### Description

change coupon discount copy in offer in trial


kkarasek (approved):
👍

PR #307, opened 2023-10-17
fix/MSSDK-1558: Fix android icon

pawelacio:
### Description

Android icon was not visible on the payment method list

- Change the svg to the colored one, because that one was a white one on the white background
- Remove `height` and `width` attributes of the svg element because when they are specified the viewBox is removed


kkarasek (approved):
👍

PR #310, opened 2023-10-19
refactor/MSSDK-1589: InnerPopupWrapper refactor

kkarasek:
### Description

This one refers to the refactor of the `InnerPopupWrapper` chunk. The component has been re-written to Typescript. 
Additionally `styled-components`' props have been replaced with [transient props](https://styled-components.com/docs/api#transient-props) to minimize the DOM pollution.

### Updates

- component rewritten to Typescript
- added styled-components transient props and updated them where necessary


src/components/InnerPopupWrapper/InnerPopupWrapper.types.ts
```diff
@@ -0,0 +1,9 @@
+import { ReactNode } from 'react';
+
+export type InnerPopupWrapperProps = {
+  steps: number;
+  popupTitle?: string;
+  currentStep: number;
+  children: ReactNode;
```
kkarasek, line 7: is it actually the correct way to type children? Also heard about the usage of React.FC<Props> that somehow has an impact on children. Any hints appreciated.

pawelacio: I think that `React.ReactNode` is the best way here. You can read something about children typing here https://frontlive.pl/blog/react-children-i-typescript

pawelacio (approved):

pawelacio (approved):

m4tewoosh (approved):

PR #333, opened 2023-12-19
feat/MSSDK-1634: On / Off switch for Cancellation Journeys

m4tewoosh:
### Description

We want to give option to turn off Cancellation Journey flow

### Updates

Added `skipAvailableFreeExtensionStep` prop that can be used with following components:
`MyAccount`, `Subscriptions`, `PlanDetails`


kkarasek (approved):
👍

PR #347, opened 2024-03-25
feat/MSSDK-1657: Add reCAPTCHA mechanism

pawelacio:
### Description

Add Google reCAPTCHA mechanism

### Updates

- small refactor and migration of the register components to TS (Register, RegisterForm)
- add new slice for settings (settingsSlice)
- add fetchSettings endpoint
- add google reCAPTCHA into register form

### Screenshots
image


src/components/RegisterPage/RegisterForm.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+const errorsInitialState = {
+  email: '',
+  password: '',
+  consents: '',
+  captcha: ''
+};
+
+export const Register = ({ onSuccess }: RegisterFormProps) => {
+  const [email, setEmail] = useState<string>('');
```
m4tewoosh, line 43: I think when we use not empty default value of state like string here or boolean in other lines, we don't have to set type in `<>` brackets because TS will infer it

m4tewoosh (dismissed):

PR #352, opened 2024-04-19
chore/MSSDK-1591: IDs for the tests in checkout component

Saddage:
### Description

Added some IDs requested by the QA


src/components/CheckoutPriceBox/CheckoutPriceBox.tsx
```diff
@@ -23,7 +22,17 @@ import {
   StyledRedeemButton
 } from './CheckoutPriceBoxStyled';
 
-const CheckoutPriceBox = ({ hideRedeemButton, isCheckout, onRedeemClick }) => {
+interface CheckoutPriceBoxProps {
```
pawelacio, line 25: We are using `type` for the Components props so if we want to be consistent you can change it. But if you think that `interface` is better for that use case we are open and we can adapt and do it in the future. #letsbuildthebestcodebase

Saddage: I think I've read somewhere that the interfaces were better, but I can't find that article anymore 🤔 I'll switch this to `type` :)

pawelacio (approved):

PR #359, opened 2024-04-26
feat/MSSDK-1334: remove redundant aria attributes; move some stuff to typescript

Saddage:
### Description

Removed some unnecessary aria attributes.
While I was at it, I also moved some code from JavaScript to TypeScript.


src/components/Capture/Capture.tsx
```diff
@@ -1,10 +1,10 @@
-import React, { useEffect, useState } from 'react';
+import { useEffect, useState } from 'react';
```
m4tewoosh, line 1: We need to remember later, when we make react-16 release to add this import on branch with 16

m4tewoosh: And in other places where this import is removed

src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,107 @@
...
+            <CheckMarkStyled
+              data-testid="checkmark"
+              $isMyAccount={isMyAccount}
+              $isRadioButton={isRadioButton}
+            />
+          )}
+        </CheckFrameStyled>
+        <ConsentDefinitionStyled
+          dangerouslySetInnerHTML={{
+            __html: `${children}${required && isMyAccount ? '*' : ''}`
```
m4tewoosh, line 87: Maybe we could pull this out and make variable with some meaning name, it would be easier to read

m4tewoosh (dismissed):

m4tewoosh:
Good job!  🚀

pawelacio (approved):

PR #368, opened 2024-05-23
fix/MSSDK-1732: do not show discount value when there's no confirmation of an applied discount

Saddage:
### Description

There was a bug that we would show a discount because taxes would be different between the country in which an offer was created and the country in which the purchase would have been made. This PR fixes this and takes into consideration the information from the backend if a discount has been applied to the current order or not.


m4tewoosh:
Looks good, but what if instead passing it as a prop we extracted `discount.applied` from redux directly in `Price.tsx`?

m4tewoosh (dismissed):

PR #370, opened 2024-06-03
feat/MSSDK-1478: Allow removing a coupon code from an order

Saddage:
### Description

Allow removing a coupon code from an order. This is done by adding a link button next to the `coupon code` text in the price parts section


package-lock.json
pawelacio, line 1: Do we need a package-lock file here?

m4tewoosh (approved):

PR #373, opened 2024-06-11
fix/MSSDK-1746: P2 TOD 2.0 - payment can't be finalized 

m4tewoosh:
### Description

After checking it looks like it’s frontend issue with checking the checkbox state. Checkbox state value returns `undefined` and blocks further payment processing. Fix checking of checkbox state, find more bulletproof way to check it.

### Updates

Used new `Checkbox` component in `PayPal.tsx` and added invisible input element to `CheckboxLegacy` for Adyen payment methods - to change checking state of checkbox to `checked` value of input element

### DIFF
--- a/src/components/Adyen/Adyen.js
+++ b/src/components/Adyen/Adyen.js
@@ -139,9 +139,9 @@ const Adyen = ({
 
     const checkbox = (
       <CheckboxLegacy
-        className={`adyen-checkout__bank-checkbox checkbox-wrapper-${methodName}`}
+        className={`adyen-checkout__bank-checkbox ${methodName}-inputWrapper`}
         checked={false}
-        id={`checkbox-${methodName}`}
+        id={`${methodName}-input`}
         onClickFn={(e, _, setIsChecked) => {
           e.target.parentElement.classList.remove(
             'adyen-checkout__bank-checkbox--error'
@@ -162,9 +162,7 @@ const Adyen = ({
         `.adyen-checkout__payment-method__details`
       );
 
-      const doesCheckboxExist = document.querySelector(
-        `#checkbox-${methodName}`
-      );
+      const doesCheckboxExist = document.querySelector(`#${methodName}-input`);
 
       if (!doesCheckboxExist) {
         const checkboxWrapper = document.createElement('div');
@@ -289,19 +287,21 @@ const Adyen = ({
     const isBancontactCard =
       selectedPaymentMethodRef?.current?.methodName === 'bancontact_card';
 
-    let checkbox = document.querySelector(`#checkbox-${methodName}`);
-    let checkboxWrapper = document.querySelector(
-      `.checkbox-wrapper-${methodName}`
-    );
+    let checkbox = document.querySelector(`#${methodName}-input`);
+    let checkboxWrapper = document.querySelector(`.${methodName}-inputWrapper`);
 
     // condition below needs to be verified when new 'scheme' is added
     if (methodName === 'scheme') {
       const schemeCheckbox = document.querySelector(
-        `#checkbox-${isBancontactCard ? 'bcmc' : 'card'}`
+        `#${isBancontactCard ? 'bcmc' : 'card'}-input`
+      );
+
+      const schemeCheckboxWrapper = document.querySelector(
+        `.${isBancontactCard ? 'bcmc' : 'card'}-inputWrapper`
       );
 
       checkbox = schemeCheckbox;
-      checkboxWrapper = schemeCheckbox.parentElement;
+      checkboxWrapper = schemeCheckboxWrapper;
     }
 
     if (!checkbox?.checked) {
--- a/src/components/Adyen/Adyen.js
+++ b/src/components/Adyen/Adyen.js
@@ -141,6 +141,7 @@ const Adyen = ({
       <CheckboxLegacy
         className={`adyen-checkout__bank-checkbox checkbox-${methodName}`}
         checked={false}
+        id={`checkbox-${methodName}`}
         onClickFn={(e, _, setIsChecked) => {
           e.target.parentElement.classList.remove(
             'adyen-checkout__bank-checkbox--error'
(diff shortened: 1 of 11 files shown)


src/components/Adyen/Adyen.js
```diff
@@ -288,17 +289,21 @@ const Adyen = ({
     const isBancontactCard =
       selectedPaymentMethodRef?.current?.methodName === 'bancontact_card';
 
-    let checkbox = document.querySelector(`.checkbox-${methodName}`);
+    let checkbox = document.querySelector(`#checkbox-${methodName}`);
+    let checkboxWrapper = document.querySelector(`.checkbox-${methodName}`);
```
Saddage:
```suggestion
    let checkboxLabel = document.querySelector(`.checkbox-${methodName}`);
```

src/components/Adyen/Adyen.js
```diff
@@ -288,17 +289,21 @@ const Adyen = ({
     const isBancontactCard =
       selectedPaymentMethodRef?.current?.methodName === 'bancontact_card';
 
-    let checkbox = document.querySelector(`.checkbox-${methodName}`);
+    let checkbox = document.querySelector(`#checkbox-${methodName}`);
```
Saddage, line 292: i think that having an ID and a CSS class with the same names, but pointing to different elements is confusing. I suggest doing something more specific, so the names kinda explain themselves
like, `id='${methodName}-input`
and the class for the label (i assume the `checkboxWrapper` points to a `<label>`) would have either a CSS class or an ID with a value like `${methodName}-inputLabel`

src/components/Adyen/Adyen.js
```diff
@@ -288,17 +289,21 @@ const Adyen = ({
...
     // condition below needs to be verified when new 'scheme' is added
     if (methodName === 'scheme') {
-      checkbox = document.querySelector(
-        `.checkbox-${isBancontactCard ? 'bcmc' : 'card'}`
+      const schemeCheckbox = document.querySelector(
+        `#checkbox-${isBancontactCard ? 'bcmc' : 'card'}`
       );
+
+      checkbox = schemeCheckbox;
+      checkboxWrapper = schemeCheckbox.parentElement;
```
Saddage, line 304: is it important what kind of a parent element it is? this might change in the future, and we won't know what happened

maybe some more specific selector would be better, like `.closest`?

src/components/CheckboxLegacy/CheckboxLegacy.js
```diff
@@ -39,14 +41,21 @@ const CheckboxLegacy = ({
...
           onClickFn(e, disabled, setIsChecked);
         }}
-        role='checkbox'
         tabIndex='-1'
         aria-checked={isChecked}
         checked={isChecked}
         aria-label={children}
         className={className}
         $disabled={disabled}
+        data-testid='checkbox-legacy'
```
Saddage, line 50: all kinds of IDs should be unique per page. with this here we might very easily have lots of the same IDs on one page. it would be good to make this more specific. it could have the `id` value

also, since we pass `id` in props, do we even need this attribute?

src/components/CheckboxLegacy/CheckboxLegacy.test.js
```diff
@@ -8,40 +8,44 @@ import CheckboxLegacy from 'components/CheckboxLegacy';
 describe('CheckboxLegacy component', () => {
   test('should render correctly without props', () => {
     render(<CheckboxLegacy />);
-    expect(screen.getByRole('checkbox')).toBeInTheDocument();
+    expect(screen.getByTestId('checkbox-legacy')).toBeInTheDocument();
```
Saddage, line 11: to expand on my comment from the previous file, if we had 2 checkboxes rendered, the tests here would fail :) that's why i think it's important to make them unique

src/components/Payment/PayPal/PayPal.tsx
```diff
@@ -35,11 +35,15 @@ const PayPal = ({
 
   const handleSubmit = () => {
     const checkbox = document.querySelector(
-      `.checkbox-paypal`
+      `#checkbox-paypal`
+    ) as HTMLInputElement;
+
+    const checkboxWrapper = document.querySelector(
```
Saddage, line 41: i think two things could be improved here:
1. let's not cast types. it takes away the help that TypeScript gives us
2. especially that you cast the type to be a `label` specifically, let's just look for the closest `label` element

src/components/Payment/PayPal/PayPal.tsx
```diff
@@ -79,21 +83,24 @@ const PayPal = ({
...
       </CopyStyled>
       <CheckboxWrapperStyled>
-        <CheckboxLegacy
-          className='adyen-checkout__bank-checkbox checkbox-paypal'
-          checked={isChecked}
-          onClickFn={(e) => {
-            e.target.parentElement.classList.remove(
+        <Checkbox
+          className='adyen-checkout__bank-checkbox checkbox-wrapper-paypal'
+          id='checkbox-paypal'
```
Saddage, line 88: i think the optional `?` is redundant here, since you passed the `| undefined` type

src/components/Payment/PayPal/PayPal.tsx
```diff
@@ -79,21 +83,24 @@ const PayPal = ({
...
+          isChecked={isChecked}
```
Saddage, line 89: i think you could just do an early return if the `event` is falsy, and then remove all the optional chainings and `if` checks :)

pawelacio (approved):

src/components/Adyen/Adyen.js
m4tewoosh: At this point this element is div (its `CheckboxLegacy`) but I will change it to label

src/components/CheckboxLegacy/CheckboxLegacy.js
m4tewoosh: Yup but currently in other places than `Adyen`  this `CheckboxLegacy` doesn't have any `id` passed as a prop, so for other tests to pass I would have to add those ids and more precisely change tests according to them. I can do it in this task but it will require more time.

src/components/Adyen/Adyen.js
m4tewoosh: Will change this wrapper (div element) to label and those selectors names

src/components/Adyen/Adyen.js
```diff
@@ -288,17 +289,23 @@ const Adyen = ({
     const isBancontactCard =
       selectedPaymentMethodRef?.current?.methodName === 'bancontact_card';
 
-    let checkbox = document.querySelector(`.checkbox-${methodName}`);
+    let checkbox = document.querySelector(`#checkbox-${methodName}`);
+    let checkboxWrapper = document.querySelector(
```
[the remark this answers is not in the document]
m4tewoosh: I have changed those classnames but still there's id and class because id goes to `input` element and `class` to its parent, I could add something like`inputId` and `labelId` but it requires more changes also in tests and maybe it's better to do it in other task than in P2

Saddage (approved):

PR #374, opened 2024-06-12
feat/MSSDK-1709: embeddable Cleeng script

Saddage:


src/scripts/constants.ts
```diff
@@ -0,0 +1 @@
+export const EMBEDDED_COMPONENT_DOMAIN = 'https://dmc9najbd8k7f.cloudfront.net';
```
m4tewoosh, line 1: If it's link to to hosted components I would change it to
HOSTED_COMPONENTS_DOMAIN

m4tewoosh (dismissed):

src/scripts/constants.ts
Saddage: good catch, thanks!

m4tewoosh (dismissed):

m4tewoosh (approved):

PR #378, opened 2024-06-18
feat/MSSDK-1639: Add phone number validation to the Capture form

m4tewoosh:
### Description

We want to have a phone number validation in the Capture form, so the users are informed about wrong phone number format even before trying to save the form data


pawelacio (dismissed):

src/components/Capture/CaptureForm/CaptureForm.tsx
```diff
@@ -22,6 +22,8 @@ import {
   CaptureError
 } from './CaptureFormStyled';
 
+const PHONE_NUMBER_REGEX = /(^[+]?([0-9][-|" "]?){4,16})$/;
```
Saddage, line 25: i think this could be placed in some `constants` file

m4tewoosh: Moved this and other regular expressions to new `regexConstants.ts` file

Saddage (approved):
nice! 🚀

PR #386, opened 2024-07-01
feat/MSSDK-1788: Update drop-in to the latest version to enable iDEAL 2.0

m4tewoosh:
### Description
In order to enable the usage of the preferred version of iDeal 2.0 payment method, we need to use latest (5.66.1) version of the Adyen Drop-in.

### Updates
Updated `adyen-web` package to 5.66.1 version

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "styled-components": ">= 5"
   },
   "dependencies": {
-    "@adyen/adyen-web": "~5.38.0",
+    "@adyen/adyen-web": "~5.66.1",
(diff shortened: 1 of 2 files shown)


Saddage (commented):
Looks good! weren't there some stuff that would stop working if we upgraded? 🤔

m4tewoosh:
> Looks good! weren't there some stuff that would stop working if we upgraded? 🤔

I have tested it making few payments, checking checkboxes and translations both ours and Adyen ones and everything looked fine, I believe more detailed testing will be in QA testing phase.

PR #394, opened 2024-07-30
feat/MSSDK-1826: Handle parent redirects from embedded components' level

m4tewoosh:
### Description

To ensure correct operation of navigation within customers' application, we need the embedded versions of our MSSDK components to support making their parents navigate to specified URLs.

### DIFF
--- a/src/components/ThankYouPage/ThankYouPage.jsx
+++ b/src/components/ThankYouPage/ThankYouPage.jsx
@@ -9,6 +9,7 @@ import Footer from 'components/Footer';
 import checkmarkIconBase from 'assets/images/checkmarkBase';
 import { getData } from 'util/appConfigHelper';
 import formatNumber from 'util/formatNumber';
+import { HOSTED_COMPONENTS_DOMAIN } from 'scripts/constants';
 
 import {
   ThankYouPageWrapperStyled,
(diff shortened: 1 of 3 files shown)


src/components/ThankYouPage/ThankYouPage.jsx
```diff
@@ -44,6 +45,15 @@ const ThankYouPage = ({ onSuccess }) => {
     return () => clearTimeout(timer);
   }, []);
 
+  const isHostedMSSDK = window.location.origin === HOSTED_COMPONENTS_DOMAIN;
```
Saddage, line 48: I'd move all of these outside of the component. There's no need to redefine everything on each render.
I'm sure about `isHostedMSSDK` and `handleTopNavigate`.
I just wanted to make sure. I don't expect this to happen, but it's better to be double-sure.
Is it possible that the value of local storage's `CLEENG_MY_ACCOUNT_URL` variable changes during runtime? If yes, let's leave it here, but if not - let's also move it outside of the component

m4tewoosh: I will move `isHostedMSSDK` and `handleTopNavigate` to new util file `hostedComponentsHelper.ts`.
When it comes to `CLEENG_MY_ACCOUNT_URL`, I will leave it in component file but move outside component

Saddage (approved):

PR #404, opened 2024-09-03
feat/MSSDK-1887: Migrate AdditionalProfileInfo, AddressDetails and useMessage to TypeScript

m4tewoosh:


pawelacio (dismissed):

src/components/AdditionalProfileInfo/AdditionalProfileInfo.tsx
```diff
@@ -15,23 +14,35 @@ import {
...
   const [isSectionDisabled, setIsSectionDisabled] = useState(true);
   const [isPending, setIsPending] = useState(false);
-  const [customSettings, setCustomSettings] = useState(null);
-  const [initialSettings, setInitialSettings] = useState(null);
-  const [message, type, setMessage, resetMessage] = useMessage();
+  const [customSettings, setCustomSettings] = useState<CustomSetting[] | null>(
+    null
+  );
+  const [initialSettings, setInitialSettings] = useState<
+    CustomSetting[] | null
```
Saddage, line 39: would an empty array make sense as the default value here?

src/components/AdditionalProfileInfo/AdditionalProfileInfo.tsx
```diff
@@ -69,37 +88,41 @@ const AdditionalProfileInfo = ({ data, isLoading, updateCaptureOption }) => {
...
-        });
-        setIsPending(false);
-        setIsSectionDisabled(true);
+
+    try {
+      await updateCaptureAnswers({
+        customAnswers
+      });
+
+      customSettings.forEach((setting) =>
```
Saddage, line 97: maybe it would make sense to have a bulk update action? what do you think? it would be just one operation instead of multiple ones

src/components/AddressDetails/AddressDetails.tsx
```diff
@@ -28,46 +50,53 @@ const AddressDetails = ({ data, isLoading, updateCaptureOption }) => {
     }
   }, [data]);
 
-  const onAddressChange = (key, newValue) => {
-    setAddress({
-      ...address,
-      [key]: newValue
-    });
-  };
+  const onAddressChange =
```
Saddage, line 53: this will get recreated on each render. please wrap this function with `useCallback`

src/components/AddressDetails/AddressDetails.tsx
```diff
@@ -28,46 +50,53 @@ const AddressDetails = ({ data, isLoading, updateCaptureOption }) => {
...
+    (key: keyof Address) => (e: React.ChangeEvent<HTMLInputElement>) => {
+      if (address) {
+        setAddress({
+          ...address,
+          [key]: e.target.value
+        });
+      }
+    };
 
   const onCancel = () => {
```
Saddage, line 63: please use the `useCallback` hook here too

src/components/AddressDetails/AddressDetails.tsx
```diff
@@ -28,46 +50,53 @@ const AddressDetails = ({ data, isLoading, updateCaptureOption }) => {
...
     setAddress(initialAddress);
     setIsSectionDisabled(true);
   };
 
-  const onSubmit = () => {
+  const onSubmit = async () => {
```
Saddage, line 68: and here too :)

src/hooks/useMessage.ts
```diff
@@ -1,10 +1,11 @@
 import { useState } from 'react';
+import { Message, MessageType, UseMessageReturnType } from 'types/hooks.types';
 
-const useMessage = () => {
+const useMessage = (): UseMessageReturnType => {
```
Saddage, line 4: I think that the array as a return value here is "difficult" to use, as we return quite a lot of stuff from this hook. IMHO, it would be better to return an object, so when using this hook, you can get only the stuff you need from it. also, typescript will make it super easy to know what is returned, so we won't need to go inside this file to see what does the hook return

src/types/hooks.types.ts
```diff
@@ -0,0 +1,13 @@
+export type MessageType = 'success' | 'error' | '';
+
+export type Message = {
+  message: string;
+  type: MessageType;
+};
+
+export type UseMessageReturnType = [
```
Saddage, line 8: if you refactor the hook to return an objet, it would make this type more readable too.

I bet if any of us comes here in like 2 weeks, and we see that the first thing that is returned is a string, or the last thing is a function that might not do anything - we won't remember what's up :)

src/components/AddressDetails/AddressDetails.tsx
m4tewoosh: Do you think there will be any benefit of using `useCallback` on such a simple function? I was told that `useCallback` also comes with a cost and should be used when the function is really complex and causes slowness because otherwise it does not give benefit but it decreases readability

src/hooks/useMessage.ts
m4tewoosh: Thats right, I left it as array because it was initially array - but it definitely should be returned as object

Saddage (approved):

PR #409, opened 2024-09-13
feat/MSSDK-1902: Migrate CancelSwitchPopup to TypeScript

m4tewoosh:


src/components/CancelSwitchPopup/CancelSwitchPopup.tsx
```diff
@@ -22,29 +23,33 @@ const CancelSwitchPopup = () => {
...
-      baseOfferPrice
-    }
-  } = useSelector((state) => state.popupManager);
+    pendingSwitchId,
+    switchDirection,
+    switchOfferTitle: untranslatedSwitchOfferTitle,
+    baseOfferTitle: untranslatedBaseOfferTitle,
+    baseOfferExpirationDate,
+    baseOfferPrice
+  } = cancelSwitch as CancelSwitch;
```
Saddage, line 39: isnt `cancelSwitch` typed? casting a type here seems like a band-aid. please take a look at the store state slice and type that correctly, so all further usages will contain proper types automatically

m4tewoosh: I casted it because it it's type in store is `CancelSwitch | null` but yea, it's not the best idea. I will return `MyAccountError` component if `cancelSwitch` is null to get rid of casting it

Saddage (approved):

Issue #319, opened 2023-11-15
Unable to use purchase or checkout component, adyen throws error


RButtsCHQ:
We're testing the new gifting functionality using the component library in a bare-bones react site. After I complete the form and hit Pay, I get this trace from Adyen.js:
```
Adyen.js:329 Uncaught (in promise) TypeError: Cannot read properties of null (reading 'type')
    at _callee3$ (Adyen.js:329:68)
    at tryCatch (regeneratorRuntime.js:45:16)
    at Generator.<anonymous> (regeneratorRuntime.js:133:17)
    at Generator.next (regeneratorRuntime.js:74:21)
    at asyncGeneratorStep (asyncToGenerator.js:3:23)
    at _next (asyncToGenerator.js:22:9)
    at asyncToGenerator.js:27:7
    at new Promise (<anonymous>)
    at Object.<anonymous> (asyncToGenerator.js:19:12)
    at Object.onSubmit (Adyen.js:361:37)
```

This only happens when trying to purchase with a credit card; PayPal is unaffected. The issue manifests with both the Purchase and Checkout Component. I am happy to provide any information you may need.

RButtsCHQ:
Here you go!


https://github.com/Cleeng/mediastore-sdk/assets/79609595/7b78842c-7bf3-4f49-af69-9832882472b5

m4tewoosh:
@RButtsCHQ Yesterday I was investigating fix for this and I believe that I found one. It's waiting to be tested and will be released as soon as possible.

m4tewoosh:
Fix released in 5.4.2 version:
https://github.com/Cleeng/mediastore-sdk/releases/tag/5.4.2

PR #420, opened 2024-10-11
feat/MSSDK-1953: Add upgrade pending info Box

m4tewoosh:
### Description

After upgrading the subscription plan from the account menu page on Web, it will not show the new plan without refreshing the page. 

### Updates

Added box with information about pending upgrade.


src/components/OfferMyAccountCard/OfferMyAccountCard.tsx
```diff
@@ -42,6 +44,9 @@ import { OfferMyAccountCardProps } from './OfferMyAccountCard.types';
 const OfferMyAccountCard = ({ offerId }: OfferMyAccountCardProps) => {
   const { t } = useTranslation();
   const dispatch = useAppDispatch();
+
+  const [switches, setSwitches] = useState<SwitchDetail[]>([]);
```
Saddage, line 48: shouldn't this be in redux? seems like all other data that is similar to this is kept in redux 🤔

src/components/OfferMyAccountCard/OfferMyAccountCard.tsx
```diff
@@ -257,44 +320,15 @@ const OfferMyAccountCard = ({ offerId }: OfferMyAccountCardProps) => {
...
           <SubBoxContentStyled>
-            <BoxTextStyled
-              dangerouslySetInnerHTML={{
-                __html: getDescription() || ''
-              }}
-            />
+            <BoxTextStyled>{getDescription()}</BoxTextStyled>
+            {isUpgradePending && (
+              <SubBoxButtonStyled onClick={() => window.location.reload()}>
+                Refresh
```
Saddage, line 326: a translation is missing here

src/components/OfferMyAccountCard/OfferMyAccountCard.tsx
m4tewoosh: This is only used in this one place so I don't see the point of throwing it into the global store, a little extra code will be needed and it may be a slight littering of the store. Only when this state will be needed elsewhere in the application would I add it to the global store.

src/components/OfferMyAccountCard/OfferMyAccountCard.tsx
m4tewoosh: Good catch, added

src/components/OfferMyAccountCard/OfferMyAccountCard.tsx
```diff
@@ -67,6 +72,15 @@ const OfferMyAccountCard = ({ offerId }: OfferMyAccountCardProps) => {
   } = useAppSelector(selectOffer);
   const priceRules = price?.rules;
 
+  const getCustomerSwitches = async () => {
```
Saddage, line 75: for code cleanliness, please move this inside the `useEffect`. in the current form, to be consistent with the rules of hooks - it should be a dependency of the `useEffect`, which would cause an infinite re-render loop

src/components/OfferMyAccountCard/OfferMyAccountCard.tsx
```diff
@@ -174,7 +193,18 @@ const OfferMyAccountCard = ({ offerId }: OfferMyAccountCardProps) => {
...
+  const getPendingUpgradeCopy = () => {
+    return t(
+      `offer-card.switch-details.upgrade-pending`,
+      `Your upgrade is pending.`
+    );
+  };
+
   const getDescription = () => {
+    if (isUpgradePending) {
+      return getPendingUpgradeCopy();
```
Saddage, line 205: i think it doesn't need to be a separate function, you can just return the `t` call here directly

src/components/OfferMyAccountCard/OfferMyAccountCard.tsx
```diff
@@ -257,44 +320,15 @@ const OfferMyAccountCard = ({ offerId }: OfferMyAccountCardProps) => {
         <SubBoxStyled>
           <IconComponent />
           <SubBoxContentStyled>
-            <BoxTextStyled
-              dangerouslySetInnerHTML={{
-                __html: getDescription() || ''
-              }}
-            />
+            <BoxTextStyled>{getDescription()}</BoxTextStyled>
```
Saddage, line 323: please extract the result of invoking `getDescription` to a variable and reuse. we don't need to rerun this function (the check in line #319)

Saddage (approved):

PR #422, opened 2024-10-15
fix/MSSDK-1942: P3 - TODtv - PayPal payment method becomes unselectable

m4tewoosh:
### Description

PayPal payment option becomes unclickable after returning from the PayPal page or stopping the PayPal redirection from loading

### Updates

Added page refresh if page was loaded from `bfcache`

### DIFF
--- a/src/components/Offer/Offer.tsx
+++ b/src/components/Offer/Offer.tsx
@@ -50,12 +50,20 @@ const Offer = ({
   }, [couponCode]);
 
   useEffect(() => {
-    // reload Payment component if it was loaded from bfcache
-    window.addEventListener('pageshow', (event) => {
+    // PayPal is unclickable after loading page from bfcache
+    // Reload Payment component if it was loaded from bfcache so new Adyen Drop-in and PayPal component are created so there are no empty states or old session data
+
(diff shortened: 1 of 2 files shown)


Saddage:
@m4tewoosh , could you provide some more information on this? Especially on why is a page refresh a fix here? I don't have the full context, but a page refresh seems to be a workaround instead of a fix 🤔

src/components/Offer/Offer.tsx
```diff
@@ -47,6 +49,15 @@ const Offer = ({
     setCoupon(couponCode);
   }, [couponCode]);
 
+  useEffect(() => {
+    // reload Payment component if it was loaded from bfcache
```
Saddage, line 53: It would be great if you could add a short note why do we need to do this. Thanks!

src/components/Offer/Offer.tsx
```diff
@@ -47,6 +49,15 @@ const Offer = ({
...
+    window.addEventListener('pageshow', (event) => {
```
Saddage, line 54: Please add the removal of the event listener on component teardown. Thanks!

Saddage (approved):

PR #425, opened 2024-10-17
feat/MSSDK-1748: Remove redux-mock-store and redux-thunk packages

m4tewoosh:
### Description

Both `redux-mock-store` and `redux-thunk packages` are used only in the unit tests. However, what they offer is already provided by the `@reduxjs/toolkit` package that we use.

### Updates

Used `configureStore` from `@reduxjs/toolkit` instead of the one from `redux-mock-store` and removed both packages


Saddage (commented):
IMHO it is better to use `renderWithProviders` from `src/util/testHelpers.tsx`. you can make the change only once there and then reuse it across the tests.

Also, please find the usages of `<Provider` in tests and also replace them with the `renderWithProviders`. This will improve our testing codebase. Thanks!

Saddage:
@m4tewoosh , please rename the PR and the `branch` to start with `chore`

PR #428, opened 2024-10-23
feat/MSSDK-1957: Pause Billing - updating My Account Copy

m4tewoosh:
### Description

Added specific copy for paused offer with new date format

### DIFF
--- a/src/styles/variables.js
+++ b/src/styles/variables.js
@@ -71,7 +71,7 @@ export const TinyFont = '11px';
 export const MicroFont = '9px';
 
 export const BoldFont = '700';
-export const MediumFontWeight = '500';
+export const SemiBoldFont = '600';
 export const LightFont = '300';
 
 export const inputTheme = {
--- a/src/styles/variables.js
+++ b/src/styles/variables.js
@@ -71,7 +71,7 @@ export const TinyFont = '11px';
 export const MicroFont = '9px';
 
 export const BoldFont = '700';
-export const MediumFontWeight = '600';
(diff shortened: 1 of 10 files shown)


src/styles/variables.js
```diff
@@ -71,7 +71,7 @@ export const TinyFont = '11px';
 export const MicroFont = '9px';
 
 export const BoldFont = '700';
-export const MediumFontWeight = '600';
+export const MediumFontWeight = '500';
```
Saddage, line 74: why was this changed? will this be consistent with our UX designs across all the components and views?

src/util/planHelper.ts
```diff
@@ -313,4 +313,14 @@ export const dateFormat = (timestamp: number, showTime = false) => {
   return resultString;
 };
 
+export const textMonthDateFormat = (timestamp: number): string => {
+  const date = new Date(timestamp * 1000);
+
+  const month = date.toLocaleString(undefined, { month: 'long' });
```
Saddage, line 319: why `undefined`? we should format this according to the user's locale

m4tewoosh: And that's what `undefined` does here, the user's locale is taken.

"If you omit this parameter, the default locale of the JavaScript runtime is used" - from the docs

src/styles/variables.js
m4tewoosh: Changed back to 600, renamed to `SemiBoldFont`

pawelacio (approved):

pawelacio (approved):

PR #437, opened 2024-12-02
feat/MSSDK-2022: Mixpanel Tracking for PayPal Payments

m4tewoosh:
### Description

To ensure `PayPal` payments are accurately tracked in `Mixpanel`, it is necessary to implement a dedicated tracking mechanism outside the `mediastore-api` (used for other payment methods), as `PayPal` utilizes a different integration. This will provide comprehensive insights into payment behaviors and improve analytics consistency across all payment methods.

### Updates

Added `ua-parser-js` package to parse user agent data the same way as in `mediastore-api`.

Added 3 tracking events related to PayPal which are sent to Mixpanel:
- Payment Action
- Payment Succeeded
- Payment Failed

### DIFF
--- a/src/package.ts
+++ b/src/package.ts
@@ -8,6 +8,7 @@ import Auth from 'services/auth';
 import store from 'appRedux/store';
 import Config, { getData } from 'util/appConfigHelper';
 import eventDispatcher, {
+  MSSDK_PURCHASE_FAILED,
   MSSDK_PURCHASE_SUCCESSFUL
 } from 'util/eventDispatcher';
 import { isPayPalExternalId } from 'util/paymentMethodHelper';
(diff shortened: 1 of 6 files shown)


src/package.ts
```diff
@@ -19,16 +24,22 @@ import PasswordReset from 'components/PasswordReset';
...
-import eventDispatcher, {
-  MSSDK_PURCHASE_SUCCESSFUL
-} from 'util/eventDispatcher';
 
 if (typeof window !== 'undefined') {
   window.onload = () => {
     const queryString = window.location.search;
     const urlParams = new URLSearchParams(queryString);
     const externalPaymentId = urlParams.get('externalPaymentId');
-    if (externalPaymentId) {
```
Saddage, line 31: i think this change introduces a bug.

right now, the check for `externalPaymentId`'s presence got removed, so we will always send the `MSSDK_PURCHASE_SUCCESSFUL` event. Even, when the `externalPaymentId` is absent

please bring back this check, but leave the new one with the Mixpanel tracking

Saddage (changes requested):

src/package.ts
m4tewoosh: No, the check is still present but now it also checks this `externalPaymentId` using `isPayPalExternalId(externalPaymentId)` function, so the check is still here and it will not send the event when `externalPaymentId` is absent.

Saddage: @m4tewoosh, right, sorry, my mistake. I didn't fully get that from the GH changes view.

However, is it possible that we have an `externalPaymentId` that is not PayPal? Like, after a 3DS redirect? If yes, then how the code works now is that it will only send the custom event for PayPal payments, and will cease to do so for others

Saddage (approved):

PR #443, opened 2024-12-18
Feat/MSSDK-2048: Add Google reCAPTCHA into Register flow

pawelacio:
### Updates


Saddage (commented):
@pawelacio, please fix the new issue that popped up in sonarcloud

src/components/RegisterPage/RegisterForm.tsx
```diff
@@ -227,12 +228,20 @@ export const RegisterForm = ({ onSuccess }: RegisterFormProps) => {
         t={t}
       />
       <Consent error={errors.consents} onChangeFn={handleConsentsChange} />
+      <ReCAPTCHA
+        ref={recaptchaRef}
+        size='invisible'
+        badge='bottomright'
+        sitekey='6Ld0A54qAAAAANJ8mLCpJAxEp0XKtJyueFmEFVaG'
```
Saddage, line 235: It would be the best to extract this to the `.env` file. in the `.env` file, please create a new variable called something like `VITE_RECAPTCHA_SITE_KEY`. then, here you can use it as follows:


```suggestion
        sitekey=import.meta.env.VITE_RECAPTCHA_SITE_KEY
```

src/components/RegisterPage/RegisterForm.tsx
```diff
@@ -43,21 +53,15 @@ export const RegisterForm = ({ onSuccess }: RegisterFormProps) => {
...
   );
 
-  useEffect(() => {
-    return () => {
-      if (timeoutId) clearTimeout(timeoutId);
-    };
-  }, []);
+  const recaptchaRef = useRef<ReCAPTCHA>(null);
 
   const handleClickShowPassword = () =>
```
Saddage, line 66: i think all of the handlers defined in this file are good candidates for extraction to a custom hook. something like `useRegisterForm` - it could return the functions, state, etc.

right now this component is very long and could use some refactoring, at least slight

src/util/validators.ts
```diff
@@ -60,3 +60,11 @@ export function validateEmailField(value: string): string {
 
   return '';
 }
+
+export function validateCaptcha(value: string | null | undefined): string {
+  const message = i18n.t(
```
Saddage, line 65: please inline this message in the return value. this way we will call the translation function only when it is actually needed, instead of calling it each time the `validateCaptcha` function is called

a small performance optimization, but they stack up over time

Saddage (approved):

.env
```diff
@@ -1 +1,2 @@
 VITE_MEDIASTORE_SDK_VERSION=$npm_package_version
+VITE_RECAPTCHA_SITE_KEY="6Ld0A54qAAAAANJ8mLCpJAxEp0XKtJyueFmEFVaG"
```
Saddage, line 2: @pawelacio, this will make captcha work on all environments. do we want that? or do we want to only enable captcha on production?

if we want to enable it only on prod, please remove the entry from this file and add it to all the `.env.{environment}` files separately. so in `.env.staging` i suggest adding `VITE_RECAPTCHA_SITE_KEY=""` - an empty string. the same for `.env.sandbox`. then, in `.env.production`, you could provide the real value to make it work only in prod

pawelacio: I think that we can merge this one like this (we are merging this to reCAPTCHA branch not to block the stage branch) and I will add that comment into one of other ticket (for example for the purchase flow). I think that maybe for the **staging** and **sandbox** we will add some testing kay that will always be valid but its something worth checking (i will try to do it after the Christmas break)

PR #444, opened 2024-12-20
[Feat/MSSDK-2050]: Add configuration method

pawelacio:
### Description

Merge this one firstly https://github.com/Cleeng/mediastore-sdk/pull/443

### Updates


src/appRedux/publisherConfigSlice.ts
```diff
@@ -27,6 +36,19 @@ export const publisherConfigSlice = createSlice({
     updatePaymentMethods: create.reducer((state, action: PayloadAction<[]>) => {
       state.paymentMethods = action.payload;
     }),
+    updateGoogleRecaptcha: create.reducer(
+      (state, action: PayloadAction<Partial<GoogleRecaptchaType>>) => {
+        console.log(action.payload);
```
Saddage, line 41: do we need these logs?

Saddage (commented):
@pawelacio, i see duplicated changes from the other PR that introduces recaptcha. I think you might need to rebase this branch here with the branch from the other PR in order to clean up the changes, so we see only the changes from this task

i just left one question about the console.logs. otherwise, all good :)

src/appRedux/publisherConfigSlice.ts
pawelacio: Meh. I forgot to remove them. Right now should be ok

src/appRedux/publisherConfigSlice.ts
```diff
@@ -13,7 +17,12 @@ export const initialState: PublisherConfigInitialState = {
   termsUrl: '',
   resetUrl: '',
   enable3DSRedirectFlow: false,
-  isPaymentCheckboxDisabled: false
+  isPaymentCheckboxDisabled: false,
+  googleRecaptcha: {
+    showCaptchaOnPurchase: false,
+    showCaptchaOnRegister: false,
+    sitekey: import.meta.env.VITE_RECAPTCHA_SITE_KEY
```
Saddage, line 24: @pawelacio , please add the respective site key to the following files with the right value for each environment (probably our site key or an empty string if we don't want to enable captcha, e.g. for staging)
`.env.production`
`.env.sandbox`
`.env.staging`

currently we don't have these values set, so the evaluation here would resolve to `undefined`

thanks!

Saddage: ah ok, closing this - i've just read all the other messages from GitHub and Chat.

PR #446, opened 2025-01-08
feat/MSSDK-2090: Add support for GCash payment method

m4tewoosh:
### Description

Implement frontend support for the GCash payment method in the MSSDK by integrating it with the Adyen Drop-in solution. This task focuses on enabling the GCash option within the Drop-in interface, ensuring it correctly interacts with the existing Adyen-connector for payment processing. The implementation should handle the display, user interaction, and appropriate response handling for GCash payments.


src/api/Customer/types/getPaymentDetails.types.ts
```diff
@@ -21,7 +21,8 @@ type PaymentDetailCardGooglePay = GeneralPaymentDetail & {
     | 'android'
     | 'ideal'
     | 'bancontact_card'
-    | 'bancontact_mobile';
+    | 'bancontact_mobile'
+    | 'gcash';
```
Saddage, line 25: I think it would be best to have just one instance of the list of the possible payment methods. please reuse the type from `appRedux/types` so we don't need to support two separate lists that describe the same thing :)

src/components/PaymentMethod/PaymentMethod.tsx
```diff
@@ -37,6 +37,7 @@ const PaymentMethod = () => {
       case 'ideal':
       case 'bancontact_card':
       case 'bancontact_mobile':
+      case 'gcash':
```
Saddage, line 40: what is the value of `paymentMethod` if it is managed by an external service?

I think the `switch` statement does not give us much value here. Please refactor this to be just a simple if that would go like:
```javascript
if (paymentMethod === 'external') { // or whatever value it has in such a case
    return (
        <Message>
            {t(
              'paymentmethod.managed-by-external-service',
              'Managed by external service'
            )}
          </Message>
    )
}

return <PaymentCard key={id} details={paymentDetail} />;
```

it would be a very quick change that will make the code easier to comprehend and also make it more performant, as there would be less actions for the compiler to make

m4tewoosh: I want to make this change and a few others related to payment methods but now I'm not sure if some cases it won't mess up, I'll do a task on improvements in payment methods integration and there I'll throw this case for checking.

Saddage: @m4tewoosh , please attach this ticket to some parent, because we will forget about it otherwise. If there's no sensible epic, please use this one: https://cleeng.atlassian.net/browse/MSSDK-1749

Saddage (approved):

PR #452, opened 2025-01-22
feat/MSSDK-2112: Add errors for google recaptcha

pawelacio:
### Description

### Updates


src/components/Payment/Payment.tsx
```diff
@@ -305,17 +306,26 @@ const Payment = ({ onPaymentComplete }: PaymentProps) => {
...
-              'payment.error.payment-method-not-supported',
-              'Payment method not supported. Try different payment method'
-            )
-          : t(
-              'payment.error.payment-not-processed',
-              'The payment has not been processed. Please, try again with a different payment method.'
-            )
-      );
+      if (code === ERROR_CODES.CAPTCHA.VERIFICATION_FAILED) {
+        setGeneralError(
```
Saddage, line 310: please refactor this a bit and create a hashmap for the errors. this way we will be able to get rid of the `if` statement and multiple `setGeneralError` calls

what i mean is to create an object like 
```javascript
const GERNERAL_ERRORS_MAP = {
    CAPTCHA_FAILED: 'payment.error.captcha-verification-failed',
    // etc.
}
```

so then here we can just call `setGeneralError(GERNERAL_ERRORS_MAP[code])`

Saddage (approved):

PR #457, opened 2025-02-06
feat/MSSDK-2126: Create Primer component without backend integration

m4tewoosh:
### Description

To introduce a `Primer` component similar to the existing `Adyen` component in the project. Import it using lazy loading. The component will import and initialize `@primer-io/checkout-web` but will not be integrated with the backend, resulting in an expected error.

### DIFF
--- a/src/components/Payment/Payment.tsx
+++ b/src/components/Payment/Payment.tsx
@@ -39,8 +39,12 @@ import PayPal from './PayPal/PayPal';
 import DropInSection from './DropInSection/DropInSection';
 import { PaymentProps } from './Payment.types';
 
-const Adyen = lazy(() => import('components/Adyen'));
-const Primer = lazy(() => import('components/Primer'));
+const Adyen = lazy(
+  () => import(/* webpackChunkName: "adyen-component" */ 'components/Adyen')
+);
+const Primer = lazy(
+  () => import(/* webpackChunkName: "primer-component" */ 'components/Primer')
+);
 
 const Payment = ({ onPaymentComplete }: PaymentProps) => {
   const { paymentMethods: publisherPaymentMethods, isPayPalHidden } =
--- a/src/components/Payment/Payment.tsx
+++ b/src/components/Payment/Payment.tsx
@@ -57,7 +57,7 @@ const Payment = ({ onPaymentComplete }: PaymentProps) => {
 
   const [generalError, setGeneralError] = useState<string>('');
 
-  const [dropInInstance, setDropInInstance] = useState<unknown | null>(null);
+  const [dropInInstance, setDropInInstance] = useState<unknown>(null);
 
   const [isActionHandlingProcessing, setIsActionHandlingProcessing] =
     useState(false);
--- a/src/components/Payment/Payment.tsx
+++ b/src/components/Payment/Payment.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState, lazy } from 'react';
+import { useEffect, useState, lazy, Suspense } from 'react';
 import { useTranslation } from 'react-i18next';
 import { getPaymentMethods, submitPayment, submitPayPalPayment } from 'api';
 import { submitPaymentWithoutDetails } from 'appRedux/paymentSlice';
@@ -341,6 +341,32 @@ const Payment = ({ onPaymentComplete }: PaymentProps) => {
     publisherPaymentMethods
   );
 
+  const shouldShowPrimer = shouldShowGatewayComponent(
+    'primer',
+    publisherPaymentMethods
+  );
+
+  const getPaymentDropIn = () => {
+    if (shouldShowPrimer) {
+      return <Primer />;
+    }
+
+    if (shouldShowAdyen) {
+      return (
+        <Adyen
+          key={adyenKey}
+          onSubmit={onAdyenSubmit}
+          selectPaymentMethod={selectPaymentMethodHandler}
+          isPayPalAvailable={shouldShowPayPal}
+          getDropIn={getDropIn}
+          onAdditionalDetails={onAdditionalDetails}
+        />
+      );
+    }
+
+    return null;
+  };
+
   const noPaymentMethods = !publisherPaymentMethods.length;
 
   const showPayPalWhenAdyenIsReady = () =>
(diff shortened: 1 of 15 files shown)


src/components/Primer/Primer.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { Primer as PrimerSDK } from '@primer-io/checkout-web';
+
+const clientToken = 'abc';
+
+const options = {
+  container: '#msd__primerWrapper'
+};
+
+const Primer = () => {
+  const createDropIn = async () => {
```
Saddage, line 11: please move this inside the `useEffect`'s callback. right now the function will be recreated on every render, versus being created only once during the component's mount

src/components/Payment/Payment.tsx
```diff
@@ -57,11 +56,8 @@ const Payment = ({ onPaymentComplete }: PaymentProps) => {
   const [isLoading, setIsLoading] = useState(false);
 
   const [generalError, setGeneralError] = useState<string>('');
-  const [adyenKey, setAdyenKey] = useState<number | null>(null);
 
-  const [dropInInstance, setDropInInstance] = useState<
-    typeof RedirectElement | null
-  >(null);
+  const [dropInInstance, setDropInInstance] = useState<unknown>(null);
```
Saddage, line 60: please provide the right type

src/components/Payment/Payment.tsx
```diff
@@ -299,7 +293,8 @@ const Payment = ({ onPaymentComplete }: PaymentProps) => {
     }
   };
 
-  const getDropIn = (drop: typeof RedirectElement) => {
+  // add correct type during adyen-web v6 migration
```
Saddage, line 296: please create a ticket for this and link it here

src/components/Payment/Payment.tsx
```diff
@@ -299,7 +293,8 @@ const Payment = ({ onPaymentComplete }: PaymentProps) => {
...
+  const getDropIn = (drop: unknown) => {
```
Saddage, line 297: we know 1 out of the 2 types we need here. please add it

src/components/Payment/Payment.tsx
```diff
@@ -334,15 +329,20 @@ const Payment = ({ onPaymentComplete }: PaymentProps) => {
...
+    onSubmit: onAdyenSubmit,
+    selectPaymentMethod: selectPaymentMethodHandler,
+    isPayPalAvailable: shouldShowPayPal,
+    getDropIn,
+    onAdditionalDetails
+  };
 
   const noPaymentMethods = !publisherPaymentMethods.length;
 
   const showPayPalWhenAdyenIsReady = () =>
```
Saddage, line 342: i don't think this needs to be a function

src/components/PaymentDropIn/PaymentDropIn.tsx
```diff
@@ -0,0 +1,36 @@
...
+const Primer = lazy(
+  () => import(/* webpackChunkName: "primer-component" */ 'components/Primer')
+);
+
+const PaymentDropIn = ({ adyenProps }: PaymentDropInProps) => {
+  const { paymentMethods: publisherPaymentMethods } = useAppSelector(
+    selectPublisherConfig
+  );
+
+  const getPaymentDropIn = () => {
```
Saddage, line 21: please extract into a custom hook

src/components/PaymentDropIn/PaymentDropIn.types.ts
```diff
@@ -0,0 +1,13 @@
+// Add correct types from adyen-web package during migration to adyen-web v6 and Adyen component to TypeScript
+type AdyenProps = {
+  isPayPalAvailable: boolean;
+  isMyAccount?: boolean;
+  onSubmit: unknown; // get type from adyen-web v6
```
Saddage, line 5: please link the migration ticket

src/components/Primer/Primer.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+const Primer = () => {
+  useEffect(() => {
+    const createDropIn = async () => {
+      await PrimerSDK.showUniversalCheckout(CLIENT_TOKEN, options);
+    };
+    createDropIn();
+  }, []);
+
+  return <div id='msd__primerWrapper' />;
```
Saddage, line 20: please use the constant, that was its purpose :)

src/components/UpdatePaymentDetailsPopup/UpdatePaymentDetailsPopup.jsx
```diff
@@ -267,6 +267,15 @@ const UpdatePaymentDetailsPopup = () => {
   const showPayPalWhenAdyenIsReady = () =>
     shouldShowAdyen ? !!dropInInstance : true;
 
+  const adyenProps = {
```
Saddage, line 270: you could inline this, no need to keep another variable in the memory

src/components/Payment/Payment.tsx
m4tewoosh: https://cleeng.atlassian.net/browse/MSSDK-2139

src/components/UpdatePaymentDetailsPopup/UpdatePaymentDetailsPopup.jsx
m4tewoosh: IMO it looks bad when inlined but sure, can inline it

src/components/Payment/Payment.tsx
m4tewoosh: Added same comment as below with ticket linked to add correct type during v6 migration

src/components/Payment/Payment.tsx
m4tewoosh: https://cleeng.atlassian.net/browse/MSSDK-2139

src/components/Payment/Payment.tsx
Saddage: we know that one of the types is `typeof RedirectElement`. please add it, so we have as much done as we can

src/hooks/usePaymentDropIn.tsx
```diff
@@ -0,0 +1,31 @@
+import { lazy } from 'react';
```
Saddage, line 1: the `src/hooks` directory is meant to hold only context-agnostic, reusable hooks. this one is very context-specific, as it is used in only one component. please move it next to the component

The general rule is that a file should be placed in a location that is the "smallest" common parent location for all the other files that use it

src/components/UpdatePaymentDetailsPopup/UpdatePaymentDetailsPopup.jsx
Saddage: i just noticed - i don't think this should be an object. this object will be re-created on each render, causing rerenders down the line
if you pass the props as separate props, the rerender will happen only if any of them changes

PR #462, opened 2025-02-13
feat/MSSDK-2097: Add props to disable Exit Survey from MSSDK components

m4tewoosh:
### Description

Implement a prop that will disable the cancellation survey - `skipCancellationSurveyStep`. This new prop will behave similar to the existing props (`skipAvailableDowngradesStep`, `skipAvailableFreeExtensionStep`). If this one prop will be added then the survey will be omitted and the unsubscribe will be called right after the button click and then after the successful request the confirmation screen will be shown.


src/components/UpdateSubscription/UpdateSubscription.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+const UpdateSubscription = ({
+  customCancellationReasons,
+  skipAvailableDowngradesStep,
+  skipAvailableFreeExtensionStep,
+  skipCancellationSurveyStep
+}: UpdateSubscriptionProps) => {
+  const { updateSubscription } = useAppSelector((state) => state.popupManager);
+
+  if (updateSubscription?.action === 'unsubscribe') {
```
Saddage, line 15: please refactor this into a `switch` statement. the `default` should rather be `return null;` instead of an empty fragment, not to create a node.

src/components/UpdateSubscription/Unsubscribe.tsx
```diff
@@ -27,6 +27,7 @@ import STEPS from './Unsubscribe.enum';
 const Unsubscribe = ({
```
Saddage, line 27: please add unit tests, as per the team's agreement

src/components/UpdateSubscription/UpdateSubscription.tsx
```diff
@@ -0,0 +1,33 @@
+import { useAppSelector } from 'appRedux/store';
+import Unsubscribe from './Unsubscribe';
+import Resubscribe from './Resubscribe';
+
+import { UpdateSubscriptionProps } from './UpdateSubscription.types';
+
+const UpdateSubscription = ({
```
Saddage, line 7: please add unit tests for this component

Saddage (approved):

src/translations/en/translations.json
```diff
@@ -606,6 +606,7 @@
   "unsubscribe-popup.cancellation.service-switch": "Switch to a different service",
   "unsubscribe-popup.cancellation.streaming-issues": "Video streaming issues",
   "unsubscribe-popup.cancellation.too-expensive": "Subscription is too expensive",
+  "unsubscribe-popup.cancellation.exit-survey-hidden": "Exit survey hidden cancellation reason",
```
pawelacio, line 609: If that one is not used, then we can remove it from here

pawelacio (dismissed):

pawelacio (approved):

PR #464, opened 2025-02-18
chore/MSSDK-2141: update packages to solve dependabot's reported vulnerabilities

Saddage:
### Description

update packages to solve dependabot's reported vulnerabilities

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,7 +119,6 @@
     "@typescript-eslint/eslint-plugin": "^5.54.0",
     "@typescript-eslint/parser": "^5.54.0",
     "@vitejs/plugin-react-swc": "^3.7.1",
-    "cross-env": "^7.0.3",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -148,8 +148,20 @@
...
     "typescript": "^4.9.5",
     "typescript-coverage-report": "^0.7.0",
-    "vite": "^5.4.8",
+    "vite": "^6.1.0",
     "vite-plugin-svgr": "^4.2.0",
-    "vitest": "^3.0.2"
+    "vitest": "^3.0.6"
+  },
+  "pnpm": {
+    "overrides": {
```
Saddage, line 156: I had to add the overrides, because there was no way to force the upgrade of these two packages by upgrading the ones that we install directly that require them to the minimal, required version

pawelacio (approved):

PR #484, opened 2025-03-14
feat/MSSDK-2127: Integrate Primer component with backend - update payment details

m4tewoosh:
### Description

To integrate the `Primer` component with the backend using `mediastore-api` endpoint. This will allow the `Primer` Drop-in to function correctly, handling payment details through API calls.

### DIFF
--- a/src/components/Primer/Primer.tsx
+++ b/src/components/Primer/Primer.tsx
@@ -1,8 +1,5 @@
 import { useEffect } from 'react';
-import {
-  Primer as PrimerSDK,
-  UniversalCheckoutOptions
-} from '@primer-io/checkout-web';
+import { Primer as PrimerSDK } from '@primer-io/checkout-web';
 import { PaymentErrorStyled } from 'components/Payment/PaymentStyled';
 import Loader from 'components/Loader';
 import { PrimerProps } from 'types/Primer.types';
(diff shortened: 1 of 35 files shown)


src/components/Primer/Primer.tsx
```diff
@@ -55,7 +47,20 @@ const Primer = ({ selectPaymentMethod, onSubmit }: PrimerProps) => {
...
+  return (
+    <>
+      <PrimerContainer id={CONTAINER} />
+      {isMyAccount && primerCheckout?.submit && (
+        <UpdateButtonStyled
+          disabled={isButtonDisabled}
+          onClick={() => primerCheckout?.submit()}
+          type='submit'
+        >
+          Update
```
Saddage, line 59: please use a translation here

src/components/Primer/Primer.tsx
```diff
@@ -55,7 +47,20 @@ const Primer = ({ selectPaymentMethod, onSubmit }: PrimerProps) => {
...
   if (isLoading) return <Loader />;
 
-  return <div id={CONTAINER} />;
+  return (
+    <>
+      <PrimerContainer id={CONTAINER} />
+      {isMyAccount && primerCheckout?.submit && (
+        <UpdateButtonStyled
+          disabled={isButtonDisabled}
+          onClick={() => primerCheckout?.submit()}
```
Saddage, line 56:
```suggestion
          onClick={primerCheckout?.submit}
```
this way you don't recreate the function on each render

Saddage: also, you can get rid of the optional chaining, as you check for the function's exitence earlier

src/components/Primer/contants.ts
```diff
@@ -0,0 +1,2 @@
+export const CONTAINER = 'msd__primerWrapper';
+export const DEFAULT_PRIMER_PAYMENT_METHOD = 'primer-card';
```
Saddage, line 2: this is a part of some type, right? please add it here, so we don't have mismatches later

src/components/Primer/contants.ts
```diff
@@ -0,0 +1,2 @@
+export const CONTAINER = 'msd__primerWrapper';
```
Saddage, line 1:
```suggestion
export const CONTAINER_CLASS_NAME = 'msd__primerWrapper';
```

also, would be good to replace `CONTAINER` with something more specific

src/components/Primer/usePrimer.ts
```diff
@@ -1,75 +1,139 @@
...
 
-// Create your own type definition to replace the non-exported one
-type PrimerClientError = {
-  code: string;
-  message: string;
-  [key: string]: any;
-};
-
-export const usePrimer = ({ onSubmit }: UsePrimerHookProps) => {
+export const usePrimer = ({ onSubmit, isMyAccount }: UsePrimerHookProps) => {
```
Saddage, line 19: please add unit tests for this hook

src/types/Primer.types.ts
```diff
@@ -1,9 +1,11 @@
 export type PrimerProps = {
-  onSubmit: () => void;
+  onSubmit?: () => void;
```
Saddage, line 2: why does this become optional? seems like it should be required all the time

m4tewoosh: It's not used in MyAccount

src/components/Primer/contants.ts
m4tewoosh: It's not a part of any type, just two constants used in both `Primer.tsx` and `usePrimer.ts` files

PR #505, opened 2025-06-02
feat/MSSDK-2456: add support to MYR currency

iwonakulacz:
### Description

### Updates

### DIFF
--- a/src/components/OfferSwitchCard/OfferSwitchCard.tsx
+++ b/src/components/OfferSwitchCard/OfferSwitchCard.tsx
@@ -48,7 +48,7 @@ const OfferSwitchCard = ({ baseOfferId, toOfferId }: OfferSwitchCardProps) => {
         CurrencyFormat,
         string
       >
-    ];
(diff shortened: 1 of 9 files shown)


src/components/OfferSwitchCard/OfferSwitchCard.tsx
```diff
@@ -48,7 +48,7 @@ const OfferSwitchCard = ({ baseOfferId, toOfferId }: OfferSwitchCardProps) => {
         CurrencyFormat,
         string
       >
-    ];
+    ] || switchDetails?.nextPaymentPriceCurrency;
```
Saddage, line 51: this might still throw an `undefined`

Saddage: @iwonakulacz yes, but what the nullish coalescing operator does is that if the `switchDetails` does not exist, the whole expression will return `undefined`. which is, what i understood, what we wanted to avoid, right?

Saddage (approved):

PR #507, opened 2025-06-04
feat/MSSDK-2460: fix showing delivery date and time in correct timezone

m4tewoosh:
### Description

Fix showing correct timezone `deliveryDate` and `deliveryTime` in gift's edit view

### DIFF
--- a/src/components/DeliveryDetails/RecipientForm/helpers.ts
+++ b/src/components/DeliveryDetails/RecipientForm/helpers.ts
@@ -5,10 +5,13 @@ const getDate = (date: number) => {
   const day = localDate.getDate().toString().padStart(2, '0');
   return `${year}-${month}-${day}`;
 };
-const getTime = (date: number) =>
-  new Date(date * 1000).toLocaleTimeString([], {
-    hour: '2-digit',
-    minute: '2-digit'
-  });
+
+const getTime = (date: number) => {
+  const localDate = new Date(date * 1000);
+  const hours = localDate.getHours().toString();
+  const minutes = localDate.getMinutes().toString();
+
+  return `${hours.padStart(2, '0')}:${minutes.padStart(2, '0')}`;
+};
 
 export { getDate, getTime };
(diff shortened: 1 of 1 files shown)


src/components/DeliveryDetails/RecipientForm/helpers.ts
```diff
@@ -1,7 +1,14 @@
-const getDate = (date: number) =>
-  new Date(date * 1000).toISOString().split('T')[0];
-
+const getDate = (date: number) => {
+  const localDate = new Date(date * 1000);
+  const year = localDate.getFullYear();
+  const month = (localDate.getMonth() + 1).toString().padStart(2, '0');
+  const day = localDate.getDate().toString().padStart(2, '0');
+  return `${year}-${month}-${day}`;
```
Saddage, line 6: i'd say it would be better to format this also with the user's locale in mind

m4tewoosh: Those methods `getFullYear()`, `getMonth()`, `getDate()` are using user's local timezone

src/components/DeliveryDetails/RecipientForm/helpers.ts
```diff
@@ -1,7 +1,14 @@
-const getDate = (date: number) =>
-  new Date(date * 1000).toISOString().split('T')[0];
-
+const getDate = (date: number) => {
+  const localDate = new Date(date * 1000);
+  const year = localDate.getFullYear();
+  const month = (localDate.getMonth() + 1).toString().padStart(2, '0');
```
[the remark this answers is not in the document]
m4tewoosh: I want this function to return `YYYY-MM-DD` format, `padStart` guarantees that both the month and day are always two digits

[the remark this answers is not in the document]
Saddage: ah, i get it now. sounds good, but also let's format with the intl locale formatter to automatically match the user's locale and how they are used to seeing dates

src/components/DeliveryDetails/RecipientForm/helpers.ts
m4tewoosh: This functions returns value passed to HTML input `type="date"` and it accepts only `YYYY-MM-DD` format. How it's  displayed in the input is dependent on the user's browser settings and their operating system's locale (regional) settings.

PR #532, opened 2026-06-22
fix/MSSDK-735: Added expiryAt guard for small numbers

MateuszGebka:
### Description

### Updates


src/components/OfferCheckoutCard/OfferDescription/OfferDescription.tsx
```diff
@@ -276,6 +277,14 @@ const OfferDescription = ({
   const renderDescriptionForPass = () => {
     const generateDescriptionForPass = () => {
       const icon = <CalendarIcon />;
+      const durationDetails = !period && getExpiresAtDurationDetails(expiresAt);
+      if (durationDetails) {
+        const { days } = durationDetails;
+        const description = t(`pass-desc.duration.day`, `{{days}}-Day pass`, {
```
iwonakulacz: is the copy approved by Anna (eg. I'm not a fan of starting the  'Day' with uppercase)? 
Please add translation key to the translation file

iwonakulacz: I don't think extracting this into a separate function is necessary. You could simplify it to something like:

```ts
const isPassForDays = !period && Number(expiresAt) < 1000;

if (isPassForDays) {
  const description = t(
    'pass-desc.duration.day',
    '{{days}}-Day pass',
    { days: expiresAt }
  );

  return { icon, description };
}
```

Since `expiresAt` is a string when its value is below `1000`, it will work correctly as the translation parameter. It just needs to be converted to a number for the comparison.

I think this approach is easier to read and understand than extracting the logic into a separate function.

iwonakulacz (approved):

